PR #10, opened 2017-07-12
[WIP] Start conversion to serde

### LINKED ISSUE #9: Move to Serde
rustc-serialize is now deprecated in favour of Serde, which hit 1.0

torkleyy:
Fixes #9 

Only serializer is implemented so far


torkleyy:
Serde handles (sized) arrays the same as tuples. But I think ron has different syntax for those, right?

kvark:
> Serde handles (sized) arrays the same as tuples. But I think ron has different syntax for those, right?

RON doesn't use fixed-size arrays, so this limitation seems fine

torkleyy:
But the problem is that `&[i32]` serializes to `[3, 4]` but `&[i32; 2]` to `(3, 4)`.

torkleyy:
Unfortunately, I won't be able to do the deserialization today. More precisely, not until Friday :(

torkleyy:
Writing a good deserializer by hand is annoying... @Xaeroxe you asked if there's anything you could help out with, got an idea how to implement it properly?

kvark:
@torkleyy We need a parser, be it either Peg/Nom/something else. I'd prefer something with less macros, all other conditions being equal.

torkleyy:
Do you have anything against LALRPOP? https://github.com/nikomatsakis/lalrpop

kvark:
@torkleyy I haven't looked at it, to be honest. The fact that Nico is behind it inspires confidence, but I'm somewhat surprised that the root doesn't have an examples folder to get a better idea.

torkleyy:
There is a link to a tutorial which contains example code.

torkleyy:
I'm not sure if we really need LALRPOP or if it is even of any use. The JSON implementation for serde does not use one as far as I can see it.

kvark:
:shipit:

PR #17, opened 2017-07-30
POM-based parsing for deserialization

kvark:


kvark:
I don't consider this complete, but the basics are in place, so it's a good point to merge.
r? @torkleyy

src/de.rs
```diff
@@ -55,154 +64,104 @@ impl StdError for Error {
...
     fn deserialize_bool<V>(self, visitor: V) -> Result<V::Value>
         where V: Visitor<'de>
     {
-        visitor.visit_bool(self.parse_bool()?)
+        println!("deserialize_bool: {:?}", self.remainder());
+        match seq(b"true").parse(&mut self.input) {
+            Ok(_) => visitor.visit_bool(true),
+            Err(_) => match seq(b"false").parse(&mut self.input) {
+                Ok(_) => visitor.visit_bool(false),
+                Err(_) => Err(Error::ExpectedBoolean)
```
torkleyy: I think we don't need `match` here.

```rust
let value = seq("true")
    .parse(&mut self.input)
    .map(true)
    .or_else(|_| seq("false")
        .map(false)
    )?;

visitor.visit_bool(value)
```

src/de.rs
```diff
@@ -305,59 +275,69 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
         visitor: V
     ) -> Result<V::Value>
         where V: Visitor<'de>
     {
-        // TODO: Should we also allow `MyUnitValue`?
-
-        self.deserialize_unit(visitor)
+        match self.consume(name) {
+            Ok(_) => visitor.visit_unit(),
+            Err(_) => Err(Error::ExpectedStructName),
```
torkleyy: I think we should also allow `()`. Not sure of how much use `MyUnitValue` is, actually.

src/de.rs
```diff
@@ -305,59 +275,69 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                Ok(_) => {
+                    let value = visitor.visit_newtype_struct(&mut *self)?;
+                    let _ = comma().parse(&mut self.input);
+                    self.consume(")")
+                        .map(|_| value)
+                        .map_err(|_| Error::ExpectedStructEnd)
+                },
+                Err(_) => Err(Error::ExpectedStruct),
+            },
+            Err(_) => visitor.visit_newtype_struct(self),
```
torkleyy: Wouldn't this mean that `Foo(9.0)` and `9.0` is permitted while `(9.0)` is not? That's somehow inconsistent, although I'm not sure what should be allowed.

src/de.rs
kvark: This code doesn't try to parse `seq("false")`, it see,s

src/de.rs
kvark: I see `()` as already being overloaded. It can be used to represent a unit, but also - an empty non-unit struct (defined as `MyStruct{}`), since we can omit struct names, and their parenthesis are `(`/`)`.

torkleyy: Still all of them don't have much meaning, it's always somethint "empty".

src/de.rs
kvark: I'd say screw it. `9.0` is inconsistent with the design, so I'd rather just support omitting the struct name as a general rule and be done with it. Hence, no `0.9`, but `(0.9)` is fine. That's what the latest code does.

src/de.rs
```diff
@@ -305,59 +276,75 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     fn deserialize_option<V>(self, visitor: V) -> Result<V::Value>
         where V: Visitor<'de>
     {
-        if self.input.starts_with("null") {
-            self.input = &self.input["null".len()..];
-            visitor.visit_none()
-        } else {
-            visitor.visit_some(self)
+        match seq(b"None").discard().parse(&mut self.input) {
```
torkleyy: Yeah, that's much better than `null`.

src/de.rs
kvark: I'm open to further discussion on the topic. It's rather minor though.

PR #18, opened 2017-07-30
Fix serialization of chars

torkleyy:
Previously, chars were serialized like strings, with `""`.


kvark:
Need tests!

torkleyy:
I don't get what's wrong :confused:

torkleyy:
Now I do, I think it's just that we cannot support deserializing into a borrowed string.
Which makes sense, because of the escaping stuff.

kvark:
Squash please?

torkleyy:
Sure, just wanted to make sure the changes appear in parts, so you don't have to check everything twice.

kvark:
Thanks, I can handle 27 lines :)

PR #23, opened 2017-07-30
Handle escapes

torkleyy:
Well, I thought it is wrong and it turns out it is. Seems there are really many things you can get wrong when writing a config format.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -717,6 +717,6 @@ mod tests {
 
     #[test]
     fn test_escape() {
-        assert_eq!("\"Quoted\"", from_str::<String>(r#"\"Quoted\""#).unwrap());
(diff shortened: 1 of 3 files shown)


src/de.rs
```diff
@@ -228,9 +230,24 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     fn deserialize_char<V>(self, visitor: V) -> Result<V::Value>
         where V: Visitor<'de>
     {
-        let parser = sym(b'\'') * take(1) - sym(b'\'');
+        let parser = sym(b'\'') * take(1);
```
torkleyy: There is most likely a better way to do this with POM, but it didn't work for me and this is somehow easier to read. I wonder if there is a parser library that is fast and nice to read.

kvark:
Good stuff!
bors r+

PR #27, opened 2017-08-01
Deserializer rework

torkleyy:
This PR gets rid of the parser (POM). The goal is to make deserialization faster and avoid unnecessary dependencies. It seems that RON is simple enough to be parsed without some framework; these libraries tend to have `Box<Fn>` and `Vec` all over the place.


kvark:
r=me when ready

kvark:
Awesome work, :shipit: !
bors r+

kvark:
sudo
bors r+

PR #29, opened 2017-08-02
Better Errors

### LINKED ISSUE #21: Parsing error locations
Current errors are not very helpful since they pin no location on where they occur. Would be nice to have something like "expected A, got B", or "expected A at line XXX col YYY".

torkleyy:
Fixes #21

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -466,12 +466,12 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
 struct CommaSeparated<'a, 'de: 'a> {
     de: &'a mut Deserializer<'de>,
     terminator: u8,
-    first: bool,
+    had_comma: bool,
 }
 
 impl<'a, 'de> CommaSeparated<'a, 'de> {
     fn new(terminator: u8, de: &'a mut Deserializer<'de>) -> Self {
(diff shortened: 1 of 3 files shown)


src/de.rs
```diff
@@ -4,14 +4,20 @@ use std::fmt;
...
+use parse::{Bytes, Position};
 
 use serde::de::{self, Deserializer as Deserializer_, DeserializeSeed, Visitor};
 
 pub type Result<T> = ::std::result::Result<T, Error>;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Error {
+    Message(String),
+    Parser { kind: ErrorKind, pos: Position },
```
kvark: nit: doesn't look like names are useful here, could just be `Parser(ErrorKind, Position)`

src/de.rs
```diff
@@ -4,14 +4,20 @@ use std::fmt;
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum ErrorKind {
```
kvark: should be `ParserError` or even `Parserror`?

torkleyy:
I have addressed your request, will merge it now because I want to continue work and start another PR. In case you find there are other changes needed, I'll address them in a future PR.

bors r+

PR #34, opened 2017-08-04
Ron `Value`s

torkleyy:
This allows for untyped deserialization and is a standard for Serde libraries like toml-rs, json, ..

### DIFF
--- a/src/de/value.rs
+++ b/src/de/value.rs
@@ -1,0 +1,223 @@
+use std::collections::BTreeMap;
+use std::cmp::Eq;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+
+use serde::de::*;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Field {
+    pub name: String,
+    pub value: Value,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Number(f64);
+
+impl Number {
+    /// Panics if `f` is not finite.
+    pub fn new(f: f64) -> Self {
+        assert!(f.is_finite());
+
+        Number(f)
+    }
+
+    pub fn into_float(self) -> f64 {
+        self.into()
+    }
+}
+
+impl Into<f64> for Number {
+    fn into(self) -> f64 {
+        self.0
+    }
+}
+
+impl Eq for Number {}
+
+impl Hash for Number {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write_i64(self.0 as i64);
+    }
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum Value {
+    Boolean(bool),
+    Byte(u8),
+    Char(char),
+    List(Vec<Value>),
+    Option(Option<Box<Value>>),
+    Map(BTreeMap<Value, Value>),
+    Number(Number),
+    String(String),
+    Struct(Option<String>, Vec<Field>),
+    Tuple(Vec<Value>),
+    Unit,
+}
+
+impl<'de> Deserialize<'de> for Value {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: Deserializer<'de>
+    {
+        deserializer.deserialize_any(ValueVisitor)
+    }
+}
+
+pub struct ValueVisitor;
+
+impl<'de> Visitor<'de> for ValueVisitor {
+    type Value = Value;
+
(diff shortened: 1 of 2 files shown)


src/de/value.rs
```diff
@@ -0,0 +1,223 @@
...
+            vec.push(elem);
+        }
+
+        Ok(Value::List(vec))
+    }
+
+    fn visit_map<A>(self, map: A) -> Result<Self::Value, A::Error>
+        where A: MapAccess<'de>,
+    {
+
```
torkleyy: Not sure how I would now whether it is a struct or a map here :thinking:

torkleyy:
Seems there is no way to implement this :(

torkleyy:
Here's what I had so far, in case anybody finds this PR and wonders what I tried:

```rust
use std::collections::BTreeMap;
use std::cmp::Eq;
use std::hash::{Hash, Hasher};

use parse::Bytes;
use super::{Deserializer, ParseError, Result};

#[derive(Clone, Debug, Eq, Hash, PartialEq)]
pub struct Field {
    pub name: String,
    pub value: Value,
}

#[derive(Clone, Copy, Debug, PartialEq)]
pub struct Number(f64);

impl Number {
    /// Panics if `f` is not finite.
    pub fn new(f: f64) -> Self {
        assert!(f.is_finite());

        Number(f)
    }

    pub fn into_float(self) -> f64 {
        self.into()
    }
}

impl Into<f64> for Number {
    fn into(self) -> f64 {
        self.0
    }
}

impl Eq for Number {}

impl Hash for Number {
    fn hash<H: Hasher>(&self, state: &mut H) {
        state.write_i64(self.0 as i64);
    }
}

#[derive(Clone, Debug, Eq, Hash, PartialEq)]
pub enum Value {
    Boolean(bool),
    Byte(u8),
    Char(char),
    List(Vec<Value>),
    Option(Option<Box<Value>>),
    Map(BTreeMap<Value, Value>),
    Number(Number),
    String(String),
    Struct(Option<String>, Vec<Field>),
    Tuple(Vec<Value>),
    Unit,
}

impl Value {
    pub fn deserialize(d: &mut Deserializer) -> Result<Self> {
        Self::deserialize_inner(&mut d.bytes)
    }

    fn deserialize_inner(bytes: &mut Bytes) -> Result<Self> {
        let peek = bytes.peek().ok_or(bytes.error(ParseError::Eof))?;

        let val = match peek {
            b'0'...b'9' => Value::Number(Number::new(bytes.float()?)),
            b'"' => Value::String(bytes.string()?.into()),
            b'\'' => Value::String(bytes.string()?.into()),
            b't' | b'f' => Value::Boolean(bytes.bool()?),
            //b'(' => Value::Tuple(bytes.tuple()?),
            //b'[' => Value::Tuple(bytes.seq()?),
            b'S' => parse_some(bytes)?,
            b'N' => parse_none(bytes)?,
            _ => unimplemented!()
        };

        Ok(val)
    }
}

fn parse_tuple_or_struct(bytes: &mut Bytes) -> Result<Value> {
    use parse::IDENT_FIRST;

    if bytes.consume("(") {
        bytes.skip_ws();

        let peek = bytes.peek().ok_or(bytes.error(ParseError::Eof))?;

        if peek == b')' {
            Ok(Value::Unit)
        } else if IDENT_FIRST.contains(&peek) {
            // Struct

            unimplemented!()
        } else {
            // Tuple

            loop {
                if !bytes.comma() || bytes.peek().ok_or(bytes.error(ParseError::Eof))? == b')' {
                    break;
                }
            }

            unimplemented!()
        }
    } else {
        bytes.err(ParseError::ExpectedArray)
    }
}

fn parse_some(bytes: &mut Bytes) -> Result<Value> {
    if bytes.consume("Some") {
        let val = Value::Option(Some(Box::new(Value::deserialize_inner(bytes)?)));

        if bytes.consume(")") {
            Ok(val)
        } else {
            bytes.err(ParseError::ExpectedOptionEnd)
        }
    } else {
        bytes.err(ParseError::ExpectedOption)
    }
}

fn parse_none(bytes: &mut Bytes) -> Result<Value> {
    if bytes.consume("None") {
        Ok(Value::Option(None))
    } else {
        bytes.err(ParseError::ExpectedOption)
    }
}
```

Problems:

* Code duplication
* No way to detect enums

kvark:
@torkleyy 
This would be nice, even if not strictly required.
Would you want to reach out to `serde` folks for a bit of support on this?

torkleyy:
Well, it's not possible to transform this into the RON data model, since `Foo(4)` could be both a tuple struct and an enum, for example.

PR #33, opened 2017-08-04
README.md update

torkleyy:

PR #37, opened 2017-08-04
Add more docs

torkleyy:


src/lib.rs
```diff
@@ -1,3 +1,60 @@
+//! RON is a simple config format which looks similar to Rust syntax.
```
kvark: I think having a block comment aka `/*! ... !*/` would be more convenient for this block

kvark:
Niice
bors r+

torkleyy:
bors is stuck..

@notriddle I cannot find anything about this PR in bors' history..is there a more detailed log?

notriddle:
That history page clearly needs timestamps. Luckily, the database should have enough info to do that (Ecto puts in updated/created columns by default). https://github.com/bors-ng/bors-ng/issues/257

I'm guessing that the crash line just before #33 is the cause of this. When it crashes like that, the queue gets thrown out; this is specifically so that bors doesn't actually get "stuck."

torkleyy:
I see, thanks!

PR #42, opened 2017-08-06
Add utility function to deserialize from a reader (trait std::io::Read)

Rhuagh:


src/de/error.rs
```diff
@@ -12,6 +13,7 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Error {
+    IOError(String),
```
torkleyy: I think guidelines say it should be `IoError`.

PR #53, opened 2017-09-10
Add Value and implement deserialize_any

### LINKED ISSUE #45: RON -> JSON converter
It should be possible to write an automatic RON->JSON converter for cases when you have a .ron file and an app that only reads .json files.

Initial discussion on gitter: https://gitter.im/ron-rs/ron?at=598b80f551915d962af79f5d

### LINKED ISSUE #48: Implement deserialize_any
It looks like RON is self-describing. It should be able to tell what type is in the data without relying on a Deserializer hint.

```rust
#[macro_use]
extern crate serde_derive;

extern crate ron;

#[derive(Deserialize, Debug)]
#[serde(untagged)]
enum Untagged {
    U8(u8),
    Bool(bool),
}

fn main() {
    println!("{}", ron::de::from_str::<u8>("99").unwrap());
    println!("{}", ron::de::from_str::<bool>("true").unwrap());

    // unhelpful panic :(
    ron::de::from_str::<Untagged>("true");
}
```

torkleyy:
Fixes #45
Fixes #48

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -90,15 +90,17 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             return visitor.visit_bool(true);
         } else if self.bytes.consume_ident("false") {
             return visitor.visit_bool(false);
-        } else if self.bytes.consume_ident("Some") {
-            return visitor.visit_some(self);
+        } else if self.bytes.check_ident("Some") {
+            return self.deserialize_option(visitor);
         } else if self.bytes.consume_ident("None") {
             return visitor.visit_none();
         } else if self.bytes.consume("()") {
             return visitor.visit_unit();
         }
 
-        let _identifier = self.bytes.identifier().is_ok();
+        if self.bytes.identifier().is_ok() {
+            self.bytes.skip_ws();
+        }
 
         match self.bytes.peek_or_eof()? {
             b'(' => self.deserialize_struct("", &[], visitor),
@@ -107,7 +109,7 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             b'0' ... b'9' | b'+' | b'-' | b'.' => self.deserialize_f64(visitor),
             b'"' => self.deserialize_string(visitor),
             b'\'' => self.deserialize_char(visitor),
-            _ => unimplemented!("TODO"),
+            other => self.bytes.err(ParseError::UnexpectedByte(other as char)),
         }
     }
 
@@ -216,6 +218,8 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         where V: Visitor<'de>
     {
         if self.bytes.consume("Some") && { self.bytes.skip_ws(); self.bytes.consume("(") } {
+            self.bytes.skip_ws();
+
             let v = visitor.visit_some(&mut *self)?;
 
             self.bytes.skip_ws();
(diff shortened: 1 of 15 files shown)


torkleyy:
@dtolnay Okay. I did it like this because `bincode` also uses the `custom` function, but I can change that. That would be a (minor) breaking change though.

dtolnay:
Oh yeah good point. That may be the same reason Bincode implemented it with `custom`. I would temporarily do it this way and file a ticket to fix it in the next major release.

Actually we already have a ticket to make it not fail in the first place... #48.

torkleyy:
Yes, that would be better. It would also fix the problem with unused values in RON files raising a `"deserialize_any not supported"`-error.

kvark:
@torkleyy CI fails, please fix.

src/de/value.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        Ok(Value::Map(res))
+    }
+
+    fn visit_enum<A>(self, data: A) -> Result<Self::Value, A::Error>
+        where A: EnumAccess<'de>
+    {
+        let (value, variant) = data.variant()?;
+
+        Ok(Value::Enum(variant, Box::new(value)))
```
torkleyy: @dtolnay I couldn't figure out how to do deal with enums here yet.

torkleyy: Never mind, there is no way I could even detect an enum, so I removed `Value::Enum`.

kvark:
hmm, do we really need to have 150+ extra lines of code to handle this case? I thought it's just a matter of failing gracefully.

torkleyy:
@kvark I changed this PR to fix #48

kvark:
thanks!
bors r+

torkleyy:
@kvark This isn't mean to be merged yet.

torkleyy:
I still need to implement `deserialize_any`.

kvark:
@torkleyy I see that you used the label to tell that it's still being worked on. I think having a big large WIP in the subject is cleaner.

torkleyy:
Got the first test implemented and passing!

src/de/mod.rs
```diff
@@ -82,10 +83,34 @@ impl<'de> Deserializer<'de> {
...
+            return visitor.visit_bool(false);
+        } else if self.bytes.check_ident("Some") {
+            return self.deserialize_option(visitor);
+        } else if self.bytes.consume_ident("None") {
+            return visitor.visit_none();
+        } else if self.bytes.consume("()") {
+            return visitor.visit_unit();
+        }
+
+        if self.bytes.identifier().is_ok() {
```
torkleyy: This is a little bit difficult.. I want to ignore this if it's e.g. the name of a struct, but if we're deserializing a struct, that's a problem (because the field names aren't strings like in JSON, but just unquoted identifiers). This would mean I need context information :(

torkleyy: @dtolnay Do you have an advice how I could solve that?

dtolnay: Here are 3 ways to store context information:

- In `self`.
- Passed as function arguments, for example save the identifier and pass it to the function responsible for deserializing a struct.
- In a new Deserializer that wraps `self` and has additional fields.

src/de/mod.rs
```diff
@@ -469,10 +469,10 @@ impl<'de, 'a> de::MapAccess<'de> for CommaSeparated<'a, 'de> {
         where K: DeserializeSeed<'de>
     {
         if self.has_element()? {
-            if self.terminator == b'}' {
-                seed.deserialize(&mut *self.de).map(Some)
+            if self.terminator == b')' {
```
kvark: why is `MapAccess` separated by `)`?

torkleyy: That's because structs are also considered maps; they share some common logic. That's just a check what it is we're deserializing in order to validate the correctness of the input.

PR #56, opened 2017-10-04
Fix bugs

torkleyy:
Turns out I actually used this pattern 3 times..

### DIFF
--- a/src/de/tests.rs
+++ b/src/de/tests.rs
@@ -159,3 +159,13 @@ fn untagged() {
 
     assert_eq!(from_str::<Untagged>("true").unwrap(), Untagged::Bool(true));
 }
+
+#[test]
+fn forgot_apostrophes() {
+    let de: Result<(i32, String)> = from_str("(4, \"Hello)");
+
+    assert!(match de {
+        Err(Error::Parser(ParseError::ExpectedStringEnd, _)) => true,
+        _ => false,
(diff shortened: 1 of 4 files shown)


src/de/tests.rs
```diff
@@ -159,3 +159,16 @@ fn untagged() {
 
     assert_eq!(from_str::<Untagged>("true").unwrap(), Untagged::Bool(true));
 }
+
+#[test]
+fn forgot_apostrophes() {
+    let de: Result<(i32, String)> = from_str("(4, \"Hello)");
+
+    assert!({
+        if let Err(Error::Parser(ParseError::ExpectedStringEnd, _)) = de {
```
kvark: match ftw!

torkleyy: I wondered why this isn't easier, but how does match help here?

Xaeroxe: (He might be referencing the concept of pattern matching?)

kvark: I just don't see the point in `if let` sugar over `match` if you take both arms and occupy more lines in the end

Xaeroxe:
Ping @kvark

It'd be really nice to get this big fix released as a patch version.

torkleyy:
@Xaeroxe that's my fault, didn't address the review earlier.

PR #61, opened 2017-10-11
Pretty configuration options

### LINKED ISSUE #58: Pretty configuration
Possible things to configure: 
  - type of indentation (N spaces, tab, etc)
  - type of newlines
  - making indentation for tuples

what else?

kvark:
Fixes #58
~~BREAKING CHANGE~~

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -75,7 +75,7 @@ struct Pretty {
 }
 
 /// Pretty serializer configuration
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct PrettyConfig {
     /// New line string
     pub new_line: String,
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -1,17 +1,12 @@
 use std::error::Error as StdError;
+use std::result::Result as StdResult;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 
 use serde::ser::{self, Serialize};
 
+#[deprecated(since="0.1.4", note="please use `to_string_pretty` with `PrettyConfig::default()` instead")]
 pub mod pretty;
-
 mod value;
-
-#[cfg(not(target_os = "windows"))]
-const NEWLINE: &str = "\n";
-
-#[cfg(target_os = "windows")]
-const NEWLINE: &str = "\r\n";
 
 /// Serializes `value` and returns it as string.
 ///
@@ -29,8 +24,21 @@ pub fn to_string<T>(value: &T) -> Result<String>
     Ok(s.output)
 }
 
+/// Serializes `value` in the recommended RON layout in a pretty way.
+pub fn to_string_pretty<T>(value: &T, config: PrettyConfig) -> Result<String>
+    where T: Serialize
+{
+    let mut s = Serializer {
+        output: String::new(),
+        pretty: Some((config, Pretty { indent: 0 })),
+        struct_names: false,
+    };
+    value.serialize(&mut s)?;
+    Ok(s.output)
+}
+
 /// Serialization result.
-pub type Result<T> = ::std::result::Result<T, Error>;
+pub type Result<T> = StdResult<T, Error>;
 
 /// Serialization error.
 #[derive(Clone, Debug, PartialEq)]
@@ -61,8 +69,33 @@ impl StdError for Error {
     }
 }
 
+/// Pretty serializer state
 struct Pretty {
     indent: usize,
+}
+
+/// Pretty serializer configuration
+#[derive(Clone, Debug)]
+pub struct PrettyConfig {
+    /// New line string
+    pub new_line: String,
+    /// Indentation string
+    pub indentor: String,
+    /// Separate tuple members with indentation
+    pub separate_tuple_members: bool,
+}
+
+impl Default for PrettyConfig {
+    fn default() -> Self {
+        PrettyConfig {
+            #[cfg(not(target_os = "windows"))]
+            new_line: "\n".to_string(),
+            #[cfg(target_os = "windows")]
+            new_line: "\r\n".to_string(),
+            indentor: "    ".to_string(),
+            separate_tuple_members: false,
+        }
+    }
 }
 
 /// The RON serializer.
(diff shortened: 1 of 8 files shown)


kvark:
Non-separated tuple members:
```yaml
(
    float: (2.180000066757202, -1.1),
    tuple: ((), false),
    map: {
        0: '1',
        3: '5',
        8: '1',
        1: '2',
    },
    nested: (
        a: "Hello from \"RON\"",
        b: 'b',
    ),
    var: A(255, ""),
)
```
Separated tuple members:
```yaml
(
    float: (
        2.180000066757202,
        -1.1,
    ),
    tuple: (
        (),
        false,
    ),
    map: {
        8: '1',
        0: '1',
        3: '5',
        1: '2',
    },
    nested: (
        a: "Hello from \"RON\"",
        b: 'b',
    ),
    var: A(
        255,
        "",
    ),
)
```

src/ser/pretty.rs
```diff
@@ -1,18 +0,0 @@
-//! Provides pretty serialization with `to_string`.
-
-use super::{Pretty, Result, Serializer};
-
-use serde::ser::Serialize;
-
-/// Serializes `value` in the recommended RON layout.
-pub fn to_string<T>(value: &T) -> Result<String>
```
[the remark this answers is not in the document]
torkleyy: Could we just add a second variant and leave this function, using `PrettyConfig::default`?

src/ser/mod.rs
```diff
@@ -61,38 +67,69 @@ impl StdError for Error {
...
 struct Pretty {
     indent: usize,
 }
 
+/// Pretty serializer configuration
+pub struct PrettyConfig {
+    /// New line string
+    pub new_line: String,
+    /// Indentation string
+    pub indentor: String,
```
torkleyy: Could theoretically use `&str` or `&[u8]` instead.

src/ser/mod.rs
```diff
@@ -61,38 +67,69 @@ impl StdError for Error {
     }
 }
 
+/// Pretty serializer state
 struct Pretty {
     indent: usize,
 }
 
+/// Pretty serializer configuration
+pub struct PrettyConfig {
```
torkleyy: Let's derive `Copy`, `Clone`, `Debug` and serde traits for this struct.

src/ser/pretty.rs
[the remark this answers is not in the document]
kvark: Not sure if it's worth it? All it takes from the user is to pass in `default()` as a parameter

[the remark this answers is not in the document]
kvark: although, you are right, by keeping the old function we can make it a non-breaking change

src/ser/mod.rs
kvark: I tried, facing the lifetime hell since the `&'a` (assuming we don't want `&'static`) starts polluting `Serializer` struct

src/ser/pretty.rs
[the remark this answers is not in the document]
torkleyy: Right, I just don't see why we should bump the version.

src/ser/mod.rs
torkleyy: Yeah, let's leave it like that then. Pretty serialization isn't something that suffers from two string allocations I guess.

src/ser/mod.rs
```diff
@@ -61,38 +67,69 @@ impl StdError for Error {
...
 pub struct Serializer {
     output: String,
-    pretty: Option<Pretty>,
+    pretty: Option<(PrettyConfig, Pretty)>,
     struct_names: bool,
 }
 
 impl Serializer {
+    fn separate_tuple_members(&self) -> bool {
+        match self.pretty {
```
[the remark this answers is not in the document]
kvark: I don't think it would look much better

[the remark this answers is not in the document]
torkleyy: It's just the style that seems to be preferred. You can also leave it as is if you prefer this style.

kvark:
Thanks for the review, please take another look!

src/ser/mod.rs
```diff
@@ -1,18 +1,13 @@
 use std::error::Error as StdError;
+use std::result::Result as StdResult;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 
 use serde::ser::{self, Serialize};
 
+#[deprecated]
```
torkleyy: I wouldn't deprecate it. Better just move the new function to that module. If you really want to, we should give an explanation.

kvark: I don't see why we need `pretty` module, it's empty and serves no solid purpose.

kvark: And what `deprecated` tells you is basically that it's going to be cut off in the next breaking release, which is what I wanted to do in the first place :)

torkleyy: Yes, but let's add an explanation that one should simply use `to_string_pretty` with a default config instead.

torkleyy: So #[deprecated(since="0.1.4", note="please use `to_string_pretty` with `PrettyConfig::default()` instead")]

kvark:
@torkleyy agreed. How about now?

src/ser/mod.rs
```diff
@@ -61,38 +69,69 @@ impl StdError for Error {
     }
 }
 
+/// Pretty serializer state
 struct Pretty {
     indent: usize,
 }
 
+/// Pretty serializer configuration
+#[derive(Clone, Debug)]
```
torkleyy: We need `Deserialize` and `Serialize`, too.

kvark:
Unfortunately, that would require serde_derive dependency in non-test configuration

torkleyy:
So what's the problem?

kvark:
@torkleyy well, one more dependency imposed for all the users. Not sure if it's worth it, considering how marginal the need to serialize the config is.

Aceeri:
@kvark serde re-exports serde_derive if you use the "derive" feature flag I believe.

kvark:
This can be done later at any point, if anyone requests the feature.

kvark:
@torkleyy added serde_derive unconditionally now

torkleyy:
Let's, finally, merge this :)

bors r+

PR #63, opened 2017-10-19
moved struct_names from Serializer to PrettyConfig

IMRaziel:


src/ser/mod.rs
```diff
@@ -83,6 +82,9 @@ pub struct PrettyConfig {
     pub indentor: String,
     /// Separate tuple members with indentation
     pub separate_tuple_members: bool,
+    /// Add struct names
+    pub struct_names: bool,
+    _phantom: PhantomData<bool>,
```
torkleyy: This should be `_dummy: ()`

src/ser/mod.rs
```diff
@@ -94,6 +96,26 @@ impl Default for PrettyConfig {
...
             indentor: "    ".to_string(),
             separate_tuple_members: false,
+            struct_names: false,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl PrettyConfig {
+    #[allow(dead_code)]
```
torkleyy: Why allow dead code and why is this function private?

IMRaziel: my bad. made it public, removed attribute

torkleyy:
So if I understand correctly, this has two drawbacks:

* more verbosity for config creation
* No struct names for unpretty serialization possible

---

I think we need to add serde attributes for our dummy field, because we don't want to serialize or deserialize it.

IMRaziel:
what about this syntax:
```rust
    pub fn new_with<F>(f: F) -> Self 
        where F: Fn(&mut Self)
    {
        let mut cfg = PrettyConfig::default();
        f(&mut cfg);
        cfg
    }

    let pretty = PrettyConfig::new_with(|x|{x.separate_tuple_members=true});

```

IMRaziel:
for control of struct names in basic `to_string` maybe add `basic` function that returns config with empty indent and newline? can use it to act `to_string_pretty` as simple `to_string`
```rust
impl PrettyConfig {
    pub fn new_with<F>(f: F) -> Self 
        where F: Fn(&mut Self)
    {
        let mut cfg = PrettyConfig::default();
        f(&mut cfg);
        cfg
    }

    pub fn basic(struct_names: bool) -> PrettyConfig{
        PrettyConfig::new_with(|x|{
            x.new_line = String::from("");
            x.indentor = String::from("");
            x.struct_names = struct_names;
        })
    }
}

```

IMRaziel:
implemented some thing i mentioned in comments above. 
`Serializer.pretty` is not `Option` now. it now makes more memory allocations, because my rust skills are not good enough to fight borrow checker, when i try return string slice for new_line and indent. OTOH, code has less if blocks for Option

IMRaziel:
not sure about last commit. in original code it was 
```rust
                self.output.pop();
                self.output.pop();

```
why were there 2 pops? 
current version seems to work fine

torkleyy:
I think one pop was for the whitespace and the other for the comma, although this was inaccurate with the pretty config.

PR #64, opened 2017-10-26
Implement `Deserializer` for `Value`

torkleyy:
Just had this idea, so I thought I could implement it.


torkleyy:
Oh right :+1:

src/value.rs
```diff
@@ -51,3 +55,151 @@ pub enum Value {
...
+                seq.reverse();
+
+                visitor.visit_seq(Seq { seq })
+            }
+            Value::Unit => visitor.visit_unit(),
+        }
+    }
+
+    forward_to_deserialize_any! {
+        bool i8 i16 i32 i64 u8 u16 u32 u64 f32 f64 char str string bytes
```
torkleyy: Hold on, integers don't work yet.

torkleyy: I'll just create another PR

PR #67, opened 2017-11-25
Sublime syntax highlighter

kvark:
It's really crappy and needs more love, but sufficient for my basic needs :) Any help improving it would be great! Example:
ron-highlight

### DIFF
--- a/highlight/RON.sublime-syntax
+++ b/highlight/RON.sublime-syntax
@@ -1,0 +1,134 @@
+%YAML 1.2
+---
+name: RON
+file_extensions:
+  - ron
+scope: source.ron
+variables:
+  ident: '[A-Za-z_][A-Za-z_0-9]*'
+contexts:
+  main:
+    - include: value
+  array:
+    - match: '\['
+      scope: punctuation.section.array.begin.ron
+      push:
+        - meta_scope: meta.structure.array.ron
+        - match: '\]'
+          scope: punctuation.section.array.end.ron
+          pop: true
+        - include: value
+        - match: ','
+          scope: punctuation.separator.array.ron
+        - match: '[^\s\]]'
+          scope: invalid.illegal.expected-array-separator.ron
+  comments:
+    - match: /\*\*(?!/)
+      scope: punctuation.definition.comment.ron
+      push:
+        - meta_scope: comment.block.documentation.ron
+        - match: \*/
+          pop: true
+    - match: /\*
+      scope: punctuation.definition.comment.ron
+      push:
+        - meta_scope: comment.block.ron
+        - match: \*/
+          pop: true
+    - match: (//).*$\n?
+      scope: comment.line.double-slash.js
+      captures:
+        1: punctuation.definition.comment.ron
+  constant:
+    - match: \b(true|false)\b
+      scope: constant.language.ron
+  number:
+    # handles integer and decimal numbers
+    - match: |-
+        (?x:         # turn on extended mode
+          -?         # an optional minus
+          (?:
+            0        # a zero
+            |        # ...or...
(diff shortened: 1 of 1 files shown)


highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,148 @@
...
+      push:
+        - meta_scope: comment.block.ron
+        - match: \*/
+          pop: true
+    - match: (//).*$\n?
+      scope: comment.line.double-slash.js
+      captures:
+        1: punctuation.definition.comment.ron
+  constant:
+    - match: \b(?:true|false)\b
```
torkleyy: What does `?:` do? Wouldn't it work without?

highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,148 @@
...
+            (?:
+              [eE]   # an e character
+              [+-]?  # followed by an option +/-
+              \d+    # followed by one or more digits
+            )?       # make exponent optional
+          )?         # make decimal portion optional
+        )
+      scope: constant.numeric.ron
+  object:
+    - match: '[A-Z][a-zA-Z]*'
```
torkleyy: I'm not sure if we want to enforce capital letters by syntax rules. Also, this doesn't allow `Ipv4Addr`, does it?

highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,148 @@
...
+  dictionary:
+    - match: '\{'
+      scope: punctuation.section.dictionary.begin.ron
+      push:
+        - meta_scope: meta.structure.dictionary.ron
+        - match: '\}'
+          scope: punctuation.section.dictionary.end.ron
+          pop: true
+        # only string keys are supported by this highlighter atm
+        - match: '"'
```
torkleyy: Couldn't this just include `value`?

highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,148 @@
...
+          scope: punctuation.definition.string.begin.ron
+          push:
+            - meta_scope: meta.structure.dictionary.key.ron string.quoted.double.ron
+            - include: inside-string
+        - include: comments
+        - match: ':'
```
torkleyy: Later this this context should be pushed as soon as a value (the key) has been parsed, right?

kvark:
Thanks for a quick look! I addressed the concerns now, hopefully ;)

highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,132 @@
...
+            (?:
+              [eE]   # an e character
+              [+-]?  # followed by an option +/-
+              \d+    # followed by one or more digits
+            )?       # make exponent optional
+          )?         # make decimal portion optional
+        )
+      scope: constant.numeric.ron
+  object:
+    - match: '[A-Z][0-9a-zA-Z]*'
```
torkleyy: This shouldn't disallow a lowercase first character.

highlight/RON.sublime-syntax
```diff
@@ -0,0 +1,132 @@
...
+      scope: entity.name.class.ron
+    - match: '\('
+      scope: punctuation.section.dictionary.begin.ron
+      push:
+        - meta_scope: meta.structure.entity.ron
+        - match: '\)'
+          scope: punctuation.section.dictionary.end.ron
+          pop: true
+        - match: '[a-z_][a-zA-Z0-9_]*'
```
torkleyy: And tbis shouldn't disallow an uppercase first character. I'd suggest you add an `ident` and use it for both the struct name and the field names.

kvark:
@torkleyy ok, third time is the charm ;)

PR #72, opened 2018-01-16
Add extension support and first extension, unwrap_newtypes

### LINKED ISSUE #69: Consider adding top-level attributes
**Syntax:** `#![attribute_name]` (like in Rust)

After reading https://ozkriff.github.io/2017-12-01--devlog.html where @ozkriff said that his configs (https://raw.githubusercontent.com/ozkriff/zemeroth_assets/master/objects.ron) aren't looking very nice because of how RON handles newtypes, I thought about adding attributes which allow to configure RON for your needs. Not only would this allow for toggles like `#![unwrap_newtypes]`, but it could also be used to deal with new features. That way, we can just enforce that you need `#![feature_name]` to enable a RON feature, such that a parser without support for that feature can return a simple and understandable error.

torkleyy:
## TODO

* [x] More tests, especially regarding error handling
* [x] discuss TODOs / design
* [x] ~~docs for extensions~~ (postponed)
* [x] add syntax to Sublime plugin

## Fixes

Fixes #69

### DIFF
--- a/highlight/RON.sublime-syntax
+++ b/highlight/RON.sublime-syntax
@@ -6,6 +6,10 @@ file_extensions:
 scope: source.ron
 contexts:
   main:
+    # extensions
+    - match: '\#\!'
+      push:
+        - include: array
     - include: value
   array:
     - match: '\['
--- a/src/de/error.rs
+++ b/src/de/error.rs
@@ -23,6 +23,8 @@ pub enum ParseError {
     Eof,
     ExpectedArray,
     ExpectedArrayEnd,
+    ExpectedAttribute,
+    ExpectedAttributeEnd,
     ExpectedBoolean,
     ExpectedComma,
     ExpectedEnum,
@@ -43,6 +45,8 @@ pub enum ParseError {
     ExpectedIdentifier,
 
     InvalidEscape,
+
+    NoSuchExtension(String),
 
     UnexpectedByte(char),
 
@@ -78,6 +82,8 @@ impl StdError for Error {
                 ParseError::Eof => "Unexpected end of file",
                 ParseError::ExpectedArray => "Expected array",
                 ParseError::ExpectedArrayEnd => "Expected end of array",
+                ParseError::ExpectedAttribute => "Expected an enable attribute",
+                ParseError::ExpectedAttributeEnd => "Expected a closing `]` after the attribute",
                 ParseError::ExpectedBoolean => "Expected boolean",
                 ParseError::ExpectedComma => "Expected comma",
                 ParseError::ExpectedEnum => "Expected enum",
--- a/src/de/error.rs
+++ b/src/de/error.rs
@@ -83,7 +83,7 @@ impl StdError for Error {
                 ParseError::ExpectedArray => "Expected array",
                 ParseError::ExpectedArrayEnd => "Expected end of array",
                 ParseError::ExpectedAttribute => "Expected an enable attribute",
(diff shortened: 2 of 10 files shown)


src/de/error.rs
```diff
@@ -78,6 +82,8 @@ impl StdError for Error {
                 ParseError::Eof => "Unexpected end of file",
                 ParseError::ExpectedArray => "Expected array",
                 ParseError::ExpectedArrayEnd => "Expected end of array",
+                ParseError::ExpectedAttribute => "Expected an enable attribute",
+                ParseError::ExpectedAttributeEnd => "Expected a closing `]` after the attribute",
```
torkleyy: Ah, this could also fail if `)` is missing. The error message should reflect that.

torkleyy:
I don't see which contains method you want me to use? The extensions are
stored in a usize.

torkleyy:
bors try

highlight/RON.sublime-syntax
```diff
@@ -6,6 +6,10 @@ file_extensions:
 scope: source.ron
 contexts:
   main:
+    # extensions
```
torkleyy: This could be done better, but I think it's sufficient. Could you try it out with your editor?

torkleyy:
r? @kvark 

Please also check the syntax file, I didn't check it!

highlight/RON.sublime-syntax
kvark: I tested it. Seems to work fine.

src/de/tests.rs
```diff
@@ -169,3 +169,43 @@ fn forgot_apostrophes() {
...
+    let de: Result<String> = from_str("#![enable(invalid)] \"Hello\"");
+
+    assert_eq!(de, err(ParseError::NoSuchExtension("invalid".to_string()), 1, 18));
+}
+
+#[test]
+fn multiple_attributes() {
+    #[derive(Debug, Deserialize, PartialEq)]
+    struct New(String);
+    let de: Result<New> = from_str("#![enable(unwrap_newtypes)] #![enable(unwrap_newtypes)] \"Hello\"");
```
kvark: this is not supported by the syntax highlighter, I think
Can we require the features to always come as a list in a single pragma? e.g. `#![enable(feature1, feature2)]`

src/parse.rs
```diff
@@ -135,6 +149,21 @@ impl<'a> Bytes<'a> {
         }
     }
 
+    fn consume_all(&mut self, all: &[&str]) -> bool {
+        all
+            .iter()
+            .map(|elem| {
+                if self.consume(elem) {
+                    self.skip_ws();
+
```
kvark: weird newline

src/de/tests.rs
torkleyy: I think it is supported? It should match against the same pattern twice.

src/parse.rs
torkleyy: Personal style, I always make a newline before the return value.

kvark:
:shipit: 
bors r+

torkleyy:
Just noticed two nits and fixed them.

bors r=kvark

Boscop:
@torkleyy The syntax highlighting in Sublime is messed up as soon as I add attributes:

image

kvark:
Could you file that on https://github.com/ron-rs/sublime-ron please?

Boscop:
Done: https://github.com/ron-rs/sublime-ron/issues/6

PR #73, opened 2018-01-17
Add grammar specification

### LINKED ISSUE #46: Grammar in Repo rather than Wiki
We should probably put the specification for the grammar, and any related stuff (like recommendations or common uses), in the repo itself instead of the wiki. This way, we can modify and discuss the spec in pull requests; and the spec will be protected from a random person deciding to hit the "edit" button.

torkleyy:
Fixes #46

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -1,0 +1,118 @@
+# RON grammar
+
+This file describes the structure of a RON file in [EBNF notation][ebnf].
+If extensions are enabled, some rules will be replaced. For that, see the
+[extensions document][exts] which describes all extensions and what they override.
+
+[ebnf]: https://en.wikipedia.org/wiki/Extended_Backus–Naur_form
+[exts]: ./extensions.md
+
+## RON file
+
+```ebnf
+RON = [extensions], ws, value, ws
+```
+
+## Whitespace and comments
+
+```ebnf
+ws = { ws_single, comment }
+ws_single = "\n" | "\t" | "\r" | " "
+comment = ["//", { no_newline }, "\n"]
+```
+
+## Commas
+
+```ebnf
+comma = ws, ",", ws
+```
+
+## Extensions
+
+```ebnf
+extensions = { "#", ws, "!", ws, "[", ws, extensions_inner, ws, "]", ws }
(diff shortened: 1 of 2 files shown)


docs/grammar.md
```diff
@@ -0,0 +1,118 @@
...
+## Commas
+
+```ebnf
+comma = ws, ",", ws
+```
+
+## Extensions
+
+```ebnf
+extensions = { "#", ws, "!", ws, "[", ws, extensions_inner, ws, "]", ws }
```
kvark: I don't see any reason to support `ws` between `#` and `!`

docs/grammar.md
```diff
@@ -0,0 +1,118 @@
...
+extensions_inner = "enable", ws, "(", extension_name, { comma, extension_name }, [comma], ws, ")" 
+```
+
+For the extension names see the [`extensions.md`][exts] document.
+
+## Value
+
+```ebnf
+value = unsigned | signed | float | string | char | bool | option | list | map | tuple | struct | enum_variant
```
kvark: I don't know much about grammar specs, but can we have something that is more formal? E.g. something like `ron.ebnf` (or any other notation) that is parse-able and validate-able.

docs/grammar.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+## Numbers
+
+```ebnf
+digit = "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
+unsigned = digit, { digit }
```
kvark: we've got to support other numerical systems (I know we don't support them atm)

docs/grammar.md
torkleyy: That's what Rust allows, too. And that's what I implemented. Should I change the implementation?

docs/grammar.md
torkleyy: I've thought about putting everything into one file and creating automated tests for our parser with it. But I don't have that much time and I think this is useful as is.

docs/grammar.md
torkleyy: Agreed

PR #77, opened 2018-01-19
Add a test

suhr:
Tests are important. So I decided to add a new one.  Unlike other tests, it's small (almost cute) and easy to understand.

The test obviously doesn't pass because ron is broken.

### DIFF
--- a/tests/vector_arrays.rs
+++ b/tests/vector_arrays.rs
@@ -1,0 +1,19 @@
+extern crate ron;
+#[macro_use]
+extern crate serde;
+
+#[derive(Deserialize, Serialize)]
+struct VectorArrays {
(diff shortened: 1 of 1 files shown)


torkleyy:
I don't think it's RON that is broken. Fixed size arrays are handled like tuples in serde.

torkleyy:
https://github.com/serde-rs/serde/issues/989

suhr:
I see. Is this an intended behaviour that should be documented?

torkleyy:
It's not really something that's specific to RON, but I guess it's surprising. Documentation would be good, but where?

torkleyy:
Closing due to inactivity.

PR #78, opened 2018-01-19
Add support for hex, oct and bin numbers

### LINKED ISSUE #76: Add hexadecimal numbers
And maybe it's a good idea to add octal and binary (`0o...` and `0b...`) notations too.

torkleyy:
Fixes #76

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,10 +1,11 @@
 use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::ops::Neg;
+use std::result::Result as StdResult;
 use std::str::{FromStr, from_utf8, from_utf8_unchecked};
 
 use de::{Error, ParseError, Result};
 
-const DIGITS: &[u8] = b"0123456789";
+const DIGITS: &[u8] = b"0123456789ABCDEFabcdef";
 const FLOAT_CHARS: &[u8] = b"0123456789.+-eE";
 const IDENT_FIRST: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_";
 const IDENT_CHAR: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz_0123456789";
@@ -276,7 +277,7 @@ impl<'a> Bytes<'a> {
     }
 
     pub fn signed_integer<T>(&mut self) -> Result<T>
-        where T: FromStr + Neg<Output=T>
+        where T: Neg<Output=T> + Num,
     {
         match self.peek_or_eof()? {
             b'+' => {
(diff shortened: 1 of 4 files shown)


src/parse.rs
```diff
@@ -4,7 +4,7 @@ use std::str::{FromStr, from_utf8, from_utf8_unchecked};
 
 use de::{Error, ParseError, Result};
 
-const DIGITS: &[u8] = b"0123456789";
+const DIGITS: &[u8] = b"0123456789ABCDEF";
```
torkleyy: Should we support lowercase characters?

kvark: Since all the language's syntaxes do, I think we should as well

src/parse.rs
```diff
@@ -344,14 +344,31 @@ impl<'a> Bytes<'a> {
         s.bytes().enumerate().all(|(i, b)| self.bytes.get(i).map(|t| *t == b).unwrap_or(false))
     }
 
-    pub fn unsigned_integer<T>(&mut self) -> Result<T> where T: FromStr {
+    pub fn unsigned_integer<T: Num>(&mut self) -> Result<T> {
+        let base = if self.bytes.get(0) == Some(&b'0') {
```
kvark: nit: let's `match self.bytes.first()` here

src/parse.rs
```diff
@@ -344,14 +344,31 @@ impl<'a> Bytes<'a> {
...
+            match self.bytes.get(1).cloned() {
```
kvark: nit: why cloning?

src/parse.rs
```diff
@@ -344,14 +344,31 @@ impl<'a> Bytes<'a> {
...
+                Some(b'x') => 16,
+                Some(b'b') => 2,
+                Some(b'o') => 8,
+                _ => 10,
```
kvark: I wonder if we should `warn!()` here

src/parse.rs
```diff
@@ -474,6 +491,25 @@ impl Extensions {
     }
 }
 
+pub trait Num: Sized {
+    fn from_str(src: &str, radix: u32) -> ::std::result::Result<Self, ()>;
```
kvark: since our `Result` collides with standard one, can we `use std::result::Result as StdResult`?

src/parse.rs
```diff
@@ -474,6 +491,25 @@ impl Extensions {
...
+}
+
+macro_rules! impl_num {
+    ($ty:ident) => {
+        impl Num for $ty {
+            fn from_str(src: &str, radix: u32) -> ::std::result::Result<Self, ()> {
```
kvark: nit: maybe we can pass-through the error (`ParseIntError`) here?

kvark:
Also, please bump the patch version.

src/parse.rs
torkleyy: So I don't have to match against references.

src/parse.rs
torkleyy: Just because the first character was a `0`, this doesn't have to be any control character (this could just be another digit). If it is an invalid character, trying to parse the integer fails, so this doesn't need special handling.

src/parse.rs
torkleyy: I cannot create the `Error` without line and column, and returning `ParseIntError` isn't better because I want to use `self.error()`.

src/parse.rs
torkleyy: I could also just do `self.peek()` I guess.

torkleyy:
patch isn't enough, the features stuff was breaking.

kvark:
hmm, wait:
  1. why was the features stuff breaking? the old RONs should still work, the new serialized RONs without features should also be finely loaded by the old parsers, no?
  2. if it is breaking, we should back-port the fix on the released branch

torkleyy:
The breaking part was changing a return type from Self to Result<Self>.

kvark:
Please squash the first 2 commits, then r=me

torkleyy:
bors r=kvark

Boscop:
Any update on this? I also need hex literals :)

PR #84, opened 2018-01-21
Add support for \x?? and improve escapes

### LINKED ISSUE #82: Serialization of Control Characters
Currently ASCII control characters like  `'\x07'` are directly serialized. I expect them to be escaped like in Rust Code.

torkleyy:
~~**This PR seems to contain a regression.**~~

Fixes #82

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,3 +1,4 @@
+use std::char::from_u32 as char_from_u32;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::ops::Neg;
 use std::result::Result as StdResult;
@@ -91,13 +92,8 @@ impl<'a> Bytes<'a> {
 
         let c = if c == b'\\' {
             let _ = self.advance(1);
-            let c = self.eat_byte()?;
 
-            if c != b'\\' && c != b'\'' {
-                return self.err(ParseError::InvalidEscape);
-            }
-
-            c as char
+            self.parse_escape()?
         } else {
             // Check where the end of the char (') is and try to
             // interpret the rest as UTF-8
@@ -209,6 +205,13 @@ impl<'a> Bytes<'a> {
                 col: self.column,
             },
         )
+    }
+
+    pub fn expect_byte(&mut self, byte: u8, error: ParseError) -> Result<()> {
+        self.eat_byte().and_then(|b| match b == byte {
+            true => Ok(()),
+            false => self.err(error),
+        })
     }
 
     /// Returns the extensions bit mask.
@@ -335,6 +338,8 @@ impl<'a> Bytes<'a> {
     }
 
     pub fn string(&mut self) -> Result<ParsedStr> {
+        use std::iter::repeat;
+
         if !self.consume("\"") {
             return self.err(ParseError::ExpectedString);
         }
@@ -359,7 +364,15 @@ impl<'a> Bytes<'a> {
 
             loop {
                 let _ = self.advance(i + 1);
-                self.parse_str_escape(&mut s)?;
+                let character = self.parse_escape()?;
+                match character.len_utf8() {
+                    1 => s.push(character as u8),
+                    len => {
+                        let start = s.len();
+                        s.extend(repeat(0).take(len));
+                        character.encode_utf8(&mut s[start..]);
+                    }
+                }
 
                 let (new_i, end_or_escape) = self.bytes
                     .iter()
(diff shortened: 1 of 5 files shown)


tests/escape.rs
```diff
@@ -0,0 +1,44 @@
...
+    let s: String = to_string(&t).unwrap();
+
+    println!("Serialized: \n\n{}\n\n", s);
+
+    assert_eq!(from_str(&s), Ok(t));
+}
+
+#[test]
+fn test_ascii_10() {
+    check_same("\u{10}".to_owned());
```
torkleyy: That's what causes the issue: the parser reads this as `\x01` instead of `\x10`.

torkleyy:
This was a stupid mistake, fixed it now :)

This is ready for review.

pyfisch:
Thanks. How about using [`char::escape_debug`](https://doc.rust-lang.org/std/primitive.char.html?search=#method.escape_debug) instead for both chars and strings.
There is also a [`u32::from_str_radix`](https://doc.rust-lang.org/std/primitive.u32.html#method.from_str_radix) to parse hexadecimal numbers.
Rust itself only [support very few special escapes](https://doc.rust-lang.org/reference/tokens.html#character-escapes) and RON should IMHO not invent new ones (or copy them from JSON).

torkleyy:
That's a good idea, I'll try it out tomorrow.

pyfisch:
@torkleyy That I like to be [able to read Unicode characters](https://play.rust-lang.org/?gist=09c795f00833601793dd21c892c44c7c&version=stable).

torkleyy:
Ah, thanks for the explanation :)

kvark:
bors delegate=pyfisch

torkleyy:
I finally finished this :) r? @pyfisch @kvark

src/parse.rs
```diff
@@ -421,82 +429,93 @@ impl<'a> Bytes<'a> {
...
+            n = n << 4;
+            let byte = self.eat_byte()?;
+            let decoded = self.decode_hex(byte)?;
+            n |= decoded;
+        }
+
+        Ok(n)
+    }
+
+    /*
```
pyfisch: Remove the commented out code.

src/ser/mod.rs
```diff
@@ -156,6 +156,12 @@ impl Serializer {
                 .extend((0..pretty.indent).map(|_| config.indentor.as_str()));
         }
     }
+
+    fn serialize_escaped_str(&mut self, value: &str) {
+        self.output += "\"";
+        self.output += &value.chars().flat_map(|c| c.escape_default()).collect::<String>();
```
pyfisch: I would avoid the string allocation and immediately push the characters to the output.

tests/escape.rs
```diff
@@ -0,0 +1,45 @@
...
+fn test_ascii_chars() {
+    (1..128).into_iter().flat_map(from_u32).for_each(check_same)
+}
+
+#[test]
+fn test_ascii_string() {
+    let s: String = (1..128).into_iter().flat_map(from_u32).collect();
+
+    check_same(s);
+}
```
pyfisch: Some test with non-ascii characters would be nice.

For example `♠` and `ß`. (They should not be escaped)

src/parse.rs
```diff
@@ -421,82 +429,93 @@ impl<'a> Bytes<'a> {
...
+                    bytes = bytes << 4;
+                    bytes |= byte as u32;
+                }
 
-                        if n2 < 0xDC00 || n2 > 0xDFFF {
-                            return self.err(ParseError::InvalidEscape);
-                        }
+                if bytes == 0 {
+                    return self.err(ParseError::InvalidEscape(
+                        "Null character (\\u{0}) not allowed",
```
pyfisch: Why is the null character not allowed? "foo\0bar" is a valid string in Rust.

src/parse.rs
```diff
@@ -421,82 +429,93 @@ impl<'a> Bytes<'a> {
...
+    }
 
     fn parse_str_escape(&mut self, store: &mut Vec<u8>) -> Result<()> {
         use std::iter::repeat;
 
         match self.eat_byte()? {
+            b'\'' => store.push(b'\''),
             b'"' => store.push(b'"'),
             b'\\' => store.push(b'\\'),
             b'b' => store.push(b'\x08'),
```
pyfisch: Some of these escapes are no standard Rust escapes.

src/ser/mod.rs
torkleyy: Err right, I was trying it out and forgot about it then. Thanks!

src/parse.rs
torkleyy: I don't think we should allow it in RON though, since that would make usage in e.g. C really difficult.

torkleyy:
Okay, addressed all comments except one (about `\0`).

pyfisch:
I don't like the current handling of NUL. Currently the string "foo\0bar" will successfully be serialized. But it will fail deserialization.

My assumption is that I can serialize any Rust data type with RON and deserialize it again and get an equivalent representation. Special casing strings with a NUL character breaks this assumption.

torkleyy:
You're right that serialization and deserialization should allow the same things. If we decide to exclude `\0` I will add a check to the serialization.

@kvark @ozkriff @Xaeroxe What do you think about this?

My main concern is that adding support for `\0` would make using RON in e.g. C really difficult. Some parsers for e.g. JSON have to use special hacks and callbacks to make `\0` inside a string work.

pyfisch:
My primary concern is that if we exclude `\0` somewhere in Servo a string with a `\0` will occur and on serialization RON panics. In this case servo will probably be forced to use a different format  without this quirk. (I *think* the current use cases for debugging in Servo and WebRender do not allow strings with `\0` but it is better to be safe)

torkleyy:
It doesn't panic, just gives you an error. And really, it doesn't make _any_ sense to include `\0` in a string. If a format supports that it's a bug, not a feature.

kvark:
C interop and serialization are somewhat orthogonal. I could see a problem with other languages implementing RON support and treating this differently. For that we should have NULL handling defined as implementation dependent. There are cases in OpenGL/Vulkan spec where they do the same: it's not undefined behavior, and it's not standard, it's a defined per implementation behavior.

TL;DR: we should support `\0` in *this* implementation.

pyfisch:
bors delegate+

Merge it after squashing.

torkleyy:
bors r=pyfisch

PR #86, opened 2018-01-24
Pretty sequence indexing

kvark:
This is super useful when inspecting serialized vectors for an ability to quicky jump to an element by index. Looks like this:
```rust
    array: [
        (),// [0]
        (),// [1]
        (),// [2]
    ],
```
Ideally, I'd like to see it like this:
```rust
    array: [
        /*0*/ (),
        /*1*/ (),
        /*2*/ (),
    ],
```
But this is blocked on supporting those types of comments.


src/ser/mod.rs
```diff
@@ -97,6 +97,7 @@ impl Default for PrettyConfig {
             new_line: "\r\n".to_string(),
             indentor: "    ".to_string(),
             separate_tuple_members: false,
+            enumerate_arrays: true,
```
ozkriff: Hmmmm, are you sure that this option should be enabled by default?

kvark: Not really. Would you not want to see the indices?

src/ser/mod.rs
```diff
@@ -418,8 +424,15 @@ impl<'a> ser::SerializeSeq for &'a mut Serializer {
         value.serialize(&mut **self)?;
         self.output += ",";
 
-        if let Some((ref config, _)) = self.pretty {
+        if let Some((ref config, ref mut pretty)) = self.pretty {
+            if config.enumerate_arrays {
+                //TODO: when /**/ comments are supported, prepend the index
+                // to an element instead of appending it.
+                // Note: this doesn't if `new_line` is not an actual new line.
```
torkleyy: I suggest we assert that `PrettyConfig::new_line` contains a `b'\n'` on creation.

src/ser/mod.rs
torkleyy: I agree with @ozkriff, I think it's too noisy for it to be the default.

kvark:
@torkleyy thanks for feedback! PTAL

torkleyy:
Thank you!

bors r+

PR #87, opened 2018-01-24
[v0.1] Implement sequence indexing

kvark:
Brother of #86 for our released branch.


kvark:
published in https://crates.io/crates/ron/0.1.6

PR #90, opened 2018-01-29
[RFC] Granular features

kvark:
The idea here is to compile the minimal amount of code for serializing stuff in Gecko, given that the build times and executable size are [concerning](https://bugzilla.mozilla.org/show_bug.cgi?id=1433910).
It sucks that Travis doesn't allow me to build a multi-dimensional matrix, any ideas are welcome :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,21 @@ rust:
   - stable
   - nightly
 branches:
-  only:
-    - staging
-    - trying
-    - master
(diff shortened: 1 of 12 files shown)


kvark:
The problem with build times in Gecko appears to be solved. This PR seems rather neat conceptually, but it makes CI to have a hard time. Given the lack of other opinions, closing.

PR #93, opened 2018-02-06
Limit the pretty depth

### LINKED ISSUE #92: Configure the depth limit for item indentation
After a particular depth, don't insert newlines any more, e.g. (from @pyfisch custom format):
```
┌ Display List
│  ├─ ClipScrollNodes
│  │  └─ ClipScrollNode { id: Some(Clip(0, PipelineId(0, 1))), parent_index: ClipScrollNodeIndex(0), clip: ClippingRegion::Empty, content_rect: TypedRect(0.0×0.0 at (0.0,0.0)), node_type: ScrollFrame(ScriptAndInputEvents) }
│  ├─ Items
│  │  ├─ PushStackingContext(StackingContext at TypedRect(1024px×1077.2333333333333px at (0px,0px)) with overflow TypedRect(1024px×1077.2333333333333px at (0px,0px)): StackingContextId(0)) StackingContext: StackingContextId(0) ClippingAndScrolling { scrolling: ClipScrollNodeIndex(0), clipping: None }
│  │  ├─ SolidColor rgba(0, 0, 0, 0) @ TypedRect(1024px×1077.2333333333333px at (0px,0px)) Rect(TypedRect(1024.0×1077.2333 at (0.0,0.0))) StackingContext: StackingContextId(0) ClippingAndScrolling { scrolling: ClipScrollNodeIndex(0), clipping: None }
```

kvark:
Fixes #92
cc @pyfisch 

Output of `cargo run --example encode`:
```yaml
(
    float: (2.18,-1.1,),
    tuple: ((),false,),
    map: {0:'1',8:'1',3:'5',1:'2',},
    nested: (a:"Hello from \"RON\"",b:'b',),
    var: A(255,"",),
    array: [(),(),(),],
)
```

Unresolved questions:
  1. should we omit the trailing comma for non-pretty output? Given that it's on the same line, it sort of makes sense. It would just affect more than just the `depth_limit` parameter.
  2. do we still want to put spaces after commas for pretty output behind the depth limit?

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -78,6 +78,8 @@ struct Pretty {
 /// Pretty serializer configuration
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct PrettyConfig {
+    /// Limit the pretty-ness up to the given depth.
+    pub depth_limit: usize,
     /// New line string
     pub new_line: String,
     /// Indentation string
@@ -91,6 +93,7 @@ pub struct PrettyConfig {
 impl Default for PrettyConfig {
     fn default() -> Self {
         PrettyConfig {
+            depth_limit: !0,
             #[cfg(not(target_os = "windows"))]
             new_line: "\n".to_string(),
             #[cfg(target_os = "windows")]
@@ -129,6 +132,13 @@ impl Serializer {
         self.output
     }
 
+    fn is_pretty(&self) -> bool {
+        match self.pretty {
+            Some((ref config, ref pretty)) => pretty.indent < config.depth_limit,
+            None => false,
+        }
+    }
+
     fn separate_tuple_members(&self) -> bool {
         self.pretty
             .as_ref()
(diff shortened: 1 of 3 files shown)


pyfisch:
> I would answer both of them with "Yes".

Agreed. Thank you for coding this.

kvark:
Addressing 1/2 in a separate WIP PR #94
Not sure how to efficiently test this, tbh. Any suggestions on how to proceed?

torkleyy:
For the test I think asserting the above output would be enough.

kvark:
@torkleyy a test has been added. Notice that the hashmap only has one element (contrary to the `encode.rs`) because the ordering is not deterministic, so I don't want unrelated surprises here later down the road.

src/ser/mod.rs
```diff
@@ -466,25 +484,25 @@ impl<'a> ser::SerializeTuple for &'a mut Serializer {
...
-                &config.new_line
-            } else {
-                " "
-            };
+        if let Some((ref config, ref pretty)) = self.pretty {
+            if pretty.indent < config.depth_limit {
+                self.output += if self.separate_tuple_members() {
+                    &config.new_line
+                } else {
+                    " "
```
torkleyy: I assume this gets fixed in the follow-up? Otherwise we don't have whitespaces between tuple fields.

PR #95, opened 2018-02-23
Re-export a serde trait

da-x:


torkleyy:
Serde has reached 1.0, so I'm not quite sure why this is necessary. I don't expect the serde version to be bumped very often.

da-x:
This is independent of `serde`'s maturity. AFAIK, if crate A specifically depends on say, version 1.0.27, and crate B specifically depends on 1.0.26, comparing `serde` types between the two constitutes a type error. 

Without this change, if I want to implement a generic function that uses `ron` to deserialize, and therefore `T: serde::de::DeserializeOwned` I must depend on the same exact `serde` version that `ron` relies on, alternatively, with this change I can use `ron`'s without worries, e.g. `T: ron::DeserializeOwned`.

torkleyy:
That's not true, Cargo will find a version that matches all requirements. Since 1.0.26 is compatible with 1.0.27, it'll just use the latter one.

da-x:
Yes, I've re-read the docs and figured it out. `serde = "1.0.26"` does not mean `serde = "=1.0.26"`, but actually `serde = "^1.0.26"`, which means latest from `1.x` is picked. Surprising for some degree (maybe it should warn?). Also, with `serde = "=1.0.26"` and `ron = "*"` cargo fails simply because 1.0.27 exists. 

I opened this, because I vaguely remember cargo being able to choose two minors for the same build rather handily - apparently it does not now. Closing for now.

torkleyy:
I think it only fails in the described case if there's a lock file.

da-x:
I see. At any case, I was not able to create a build with both `1.0.27` and `1.0.26` as distinct dependencies. I was only able to have `0.*` and `1.*` trees as distinct, non conflicting dependencies.

PR #96, opened 2018-03-11
Update specification link.

memoryruins:


torkleyy:
The full grammar in EBNF syntax is available here: https://github.com/ron-rs/ron/blob/master/docs/grammar.md

PR #108, opened 2018-05-20
Allow ws in newtype variants

### LINKED ISSUE #107: `to_string` produces text that `from_str` does not accept (with tuple variants)
```
# Cargo.toml
ron = "0.2.1"
```

---

Given the following type:

```rust
pub enum FileOrMem {
    File(String),
    Memory,
}
```

`ser::to_string` produces the following RON:

```
            data: Node (
                location: Location (
                    file_or_mem: File(
                        "foo/bar"
                    ),
                    column: 1,
                    line: 1
                ),
                key: "Player",
                value: None,
                comment: None
            )
```

Which fails deserialization with:

```
Parser(ExpectedString, Position { col: 39, line: 19 })
```

The fix, IME, is to change the text to be inline. But this should be fixed in parsing IMO (what I mean is that this syntax is reasonable and should be accepted).

---

For the sake of completeness this error is not restricted to string literals on a new line.

It also occurs if there is a space between the `(` and `"`.

torkleyy:
Fixes #107

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ron"
-version = "0.2.1"
+version = "0.2.2"
 license = "MIT/Apache-2.0"
 keywords = ["parser", "serde", "serialization"]
 authors = [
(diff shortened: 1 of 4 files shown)


Phrohdoh:
AFAICT this will fix #107.

torkleyy:
Yep, added it to the PR body :)

PR #109, opened 2018-05-31
Use base64 for `serde_bytes` annotated fields

fengalin:
With the following `struct`, `small` will be represented as an array of `u8` and `large` as a base64 `String`.

``` rust
extern crate serde_derive;

extern crate serde;
extern crate serde_bytes;

#[derive(Serialize, Deserialize)]
struct BytesStruct {
    small: Vec<u8>,
    #[serde(with = "serde_bytes")]
    large: Vec<u8>,
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,10 @@ exclude = ["bors.toml", ".travis.yml"]
 name = "ron"
 
 [dependencies]
+base64 = "^0.9.2"
 bitflags = "1"
 serde = { version = "1", features = ["serde_derive"] }
 
 [dev-dependencies]
+serde_bytes = "0.10"
 serde_json = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ron"
-version = "0.2.2"
+version = "0.3.0"
 license = "MIT/Apache-2.0"
 keywords = ["parser", "serde", "serialization"]
 authors = [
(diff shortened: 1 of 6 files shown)


fengalin:
I'm glad you find this PR interesting! Is there anything I could do to help push this forward?

kvark:
@fengalin sorry about the wait. Please rebase in the meantime :)
@torkleyy must have just missed this. PING!

kvark:
oh I see the response now, oops, sorry @torkleyy  :)
@fengalin please bump the minor version and rebase

fengalin:
@kvark no problem, I was just checking.
Rebased and bumped version.

fengalin:
@torjkleyy I'm not quite sure about this. I think `base64` handling should only be selected when user explicitly uses `Bytes/_buf`.  I'll take a deeper look at this tomorrow.

torkleyy:
The problem is that even though there's a `Value::Bytes` variant, it will
never be the result of a deserialization.

fengalin:
I marked the `Value::Bytes` variant as `unreachable!` and implemented `deserialize_bytes` and `deserialize_byte_buf`. Does it address your concerns @torkleyy?

fengalin:
AFAIU, you will get the same value if the origin and target `struct` are annotated with `serde_bytes` or use `ByteBuf`. Are you saying we should use a specific notation in the serialized field to show this is base64?

torkleyy:
Yes, with fields this holds, but not with `ron::Value`.

torkleyy:
@kvark Do you still approve the PR with the new changes?

kvark:
@fengalin thanks for your changes!
@torkleyy thanks for reviewing!
bors r=torkleyy

PR #112, opened 2018-06-05
Add support for parsing raw string literals

### LINKED ISSUE #111: Support for multi-line string literals
Will RON ever support multi-line string literals? This is a feature that missing from JSON, but is present in YAML, TOML, and in Rust proper, like so:

```yaml
# YAML multi-line string literals
key: >
  This is a very long sentence
  that spans several lines in the YAML
  but which will be rendered as a string
  without carriage returns.
```

```toml
# TOML multi-line string literals
key = '''
    This is a very long sentence
    that spans several lines in the TOML
    but which will be rendered as a string
    without carriage returns.
'''
```

```rust
// Rust multi-line string literals
let key = r#"
    This is a very long sentence
    that spans several lines in Rust
    but which will be rendered as a string
    without carriage returns.
"#;

let key = r##"
    This is a sentence which happens
    to contain a '#' character inside.
"##;
```

This could be a useful for storing localized strings for dialogue boxes, for example, or other complex strings that are too tedious and error-prone to escape by hand. Here is some pseudo-RON syntax showing how this feature could be used:

```ron
System(
    arch: "x86_64-linux",
    shell: "/usr/bin/bash",
    packages: ["firefox", "git", "rustup"],
    startup: r##"
        #!/usr/bin/bash
        export PS1="..."
        echo "Done booting!"
    "##,
)
```

Given that RON already mimics Rust's notation quite well, adopting its syntax for raw string literals would make this feature feel right at home. See the [official reference page](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) and [Rust By Example](https://rustbyexample.com/std/str.html#literals-and-escapes) for details.

ebkalderon:
### Added

* Add support for parsing Rust-style raw string literals.

### Changed

* Update string unit test to verify string literal deserialization.

Fixes #111.

### DIFF
--- a/src/de/tests.rs
+++ b/src/de/tests.rs
@@ -93,8 +93,16 @@ fn test_map() {
 #[test]
 fn test_string() {
     let s: String = from_str("\"String\"").unwrap();
-
     assert_eq!("String", s);
+
+    let raw: String = from_str("r\"String\"").unwrap();
+    assert_eq!("String", raw);
+
+    let raw_hashes: String = from_str("r#\"String\"#").unwrap();
+    assert_eq!("String", raw_hashes);
+
+    let raw_hashes_quote: String = from_str("r##\"String with \"#\"##").unwrap();
+    assert_eq!("String with \"#", raw_hashes_quote);
 }
 
 #[test]
--- a/src/de/tests.rs
+++ b/src/de/tests.rs
@@ -101,6 +101,9 @@ fn test_string() {
     let raw_hashes: String = from_str("r#\"String\"#").unwrap();
     assert_eq!("String", raw_hashes);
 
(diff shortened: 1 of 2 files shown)


src/de/tests.rs
```diff
@@ -93,8 +93,16 @@ fn test_map() {
...
-
     assert_eq!("String", s);
+
+    let raw: String = from_str("r\"String\"").unwrap();
+    assert_eq!("String", raw);
+
+    let raw_hashes: String = from_str("r#\"String\"#").unwrap();
+    assert_eq!("String", raw_hashes);
+
+    let raw_hashes_quote: String = from_str("r##\"String with \"#\"##").unwrap();
```
kvark: can we test the multi-line raw strings?

ebkalderon: Sure, I'll add that in!

PR #114, opened 2018-06-06
Add 'r' byte to deserialize_any()

### LINKED ISSUE #113: Forgot to hook up raw string literals with deserialize_any
I made a mistake in #112 where I forgot to update [the appropriate branch](https://github.com/ron-rs/ron/blob/master/src/de/mod.rs#L125) in `de/mod.rs` to work with raw string literals. The match pattern should be `b'"' | b'r'` in order for raw string literal deserialization to work in `deserialize_any`. I will prepare a PR to fix this.

ebkalderon:
### Fixed

* Update pattern in `deserialize_any()` to accommodate for raw string literals.

Fixes #113.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -277,9 +277,15 @@ impl<'a> Bytes<'a> {
     }
 
     pub fn identifier(&mut self) -> Result<&'a [u8]> {
-        if IDENT_FIRST.contains(&self.peek_or_eof()?) {
-            let bytes = self.next_bytes_contained_in(IDENT_CHAR);
+        let next = self.peek_or_eof()?;
+        if IDENT_FIRST.contains(&next) {
+            // If the next two bytes signify the start of a raw string literal,
+            // return an error.
+            if next == b'r' && (self.bytes[1] == b'"' || self.bytes[1] == b'#') {
+                return self.err(ParseError::ExpectedIdentifier);
+            }
 
+            let bytes = self.next_bytes_contained_in(IDENT_CHAR);
             let ident = &self.bytes[..bytes];
             let _ = self.advance(bytes);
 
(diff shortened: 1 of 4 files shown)


ebkalderon:
@torkleyy Sure, I can do that.

ebkalderon:
Okay, I've added the integration test. In the process of doing so, however, it seems that the changes needed to accommodate raw string literals were more involved than simply adding another match pattern.

I had an issue in `deserialize_any()` where `identifier()` was being triggered accidentally due to the first byte of a raw string literal being "r". To fix this, I [updated the method](https://github.com/ebkalderon/ron/blob/raw-string-literals/src/parse.rs#L279-L296) to branch when a `b'r'` is detected to check whether the next byte is a `b'"'` or a `b'#'` before declaring the rest an identifier.

ebkalderon:
@torkleyy Everything should be up-to-date with the latest `master` now.

src/parse.rs
```diff
@@ -277,9 +277,18 @@ impl<'a> Bytes<'a> {
     }
 
     pub fn identifier(&mut self) -> Result<&'a [u8]> {
-        if IDENT_FIRST.contains(&self.peek_or_eof()?) {
-            let bytes = self.next_bytes_contained_in(IDENT_CHAR);
+        let next = self.peek_or_eof()?;
+        if IDENT_FIRST.contains(&next) {
+            // If the next two bytes signify the start of a raw string literal,
+            // return an error.
+            if next == b'r' {
```
torkleyy: I'm not quite sure why we need this extra check here, sorry.

torkleyy: Forgot to delete this comment, it makes sense with the current code, but I'm not sure it's the right way.

ebkalderon: @torkleyy The only reason I added this branch is to only perform the second byte if the next byte is an `r`. However, I agree that it may not be the best way. This was the best solution I could come up with, however, given the current code structure. If you have any better ideas, please let me know and I'll implement them!

ebkalderon:
@torkleyy It appears that neither `deserialize_enum()` nor `deserialize_identifier()` is handled by the pre-existing `deserialize_any()`. However, I think that this change may need to be made in a separate pull request.

PR #120, opened 2018-07-30
Reexport `Value` at  root

matklad:
Hi! As a client of `ron`, I'd love to be able to use `ron::Value` :0)


kvark:
Thank you!
I think we should hide `value` module then in the next breaking change.
bors r+

torkleyy:
Thanks @matklad! Seems bors is stuck?

torkleyy:
Just took a look at the log, it crashed: https://app.bors.tech/repositories/865/log

PR #124, opened 2018-08-05
Handle tuple deserialization in deserialize_any properly

### LINKED ISSUE #121: Ron does not properly round trip through Value
```rust
#[macro_use]
extern crate serde;

extern crate ron;
use ron::de::*;
use ron::ser;
use ron::value::Value;

#[derive(Serialize)]
struct Scene(Option<(u32, u32)>);

#[derive(Serialize)]
struct Scene2 {
    foo: Option<(u32, u32)>,
}

fn main() {
    {
        let s = ser::to_string(&Scene2 { foo: Some((1, 1)) }).unwrap();
        println!("{}", s);
        let scene: Value = from_str(&s).unwrap();
        println!("{:?}", scene);
    }
    {
        let s = ser::to_string(&Scene( Some((1, 1)) )).unwrap();
        println!("{}", s);
        let scene: Value = from_str(&s).unwrap();
        println!("{:?}", scene);
    }
}
```

The first example gives:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parser(ExpectedIdentifier, Position { col: 12, line: 1 })', libcore/result.rs:945:5
```

The second gives:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parser(ExpectedMapColon, Position { col: 6, line: 1 })', libcore/result.rs:945:5
```

torkleyy:
Fixes #121 

Needs some more tests before merging. I'm not happy about all the lookahead that this requires, but I don't see a better solution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ron"
-version = "0.3.0"
+version = "0.4.0"
 license = "MIT/Apache-2.0"
 keywords = ["parser", "serde", "serialization"]
 authors = [
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -92,6 +92,32 @@ impl<'de> Deserializer<'de> {
             self.bytes.err(ParseError::TrailingCharacters)
         }
     }
+
+    /// Called from `deserialze_any` when a struct was detected. Decides if there is a
+    /// unit, tuple or usual struct and deserializes it accordingly.
+    ///
+    /// This method assumes there is no identifier left.
+    fn handle_any_struct<V>(&mut self, visitor: V) -> Result<V::Value>
+    where
+        V: Visitor<'de>
+    {
+        // Create a working copy
+        let mut bytes = self.bytes.clone();
+
+        match bytes.consume("(") {
+            true => {
+                bytes.skip_ws()?;
+
+                match bytes.check_tuple_struct()? {
+                    // first argument is technically incorrect, but ignored anyway
+                    true => self.deserialize_tuple(0, visitor),
+                    // first two arguments are technically incorrect, but ignored anyway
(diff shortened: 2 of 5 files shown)


torkleyy:
Okay, let's see if bors works now.

bors try

torkleyy:
Okay, now only one issue remains: The conversion from a serialized tuple to a `Value` loses the tuple information (it becomes a sequence). That's because Serde's data model doesn't differentiate between those two.

src/de/mod.rs
```diff
@@ -92,6 +92,32 @@ impl<'de> Deserializer<'de> {
...
+    /// Called from `deserialze_any` when a struct was detected. Decides if there is a
+    /// unit, tuple or usual struct and deserializes it accordingly.
+    ///
+    /// This method assumes there is no identifier left.
+    fn handle_any_struct<V>(&mut self, visitor: V) -> Result<V::Value>
+    where
+        V: Visitor<'de>
+    {
+        // Create a working copy
+        let mut bytes = self.bytes.clone();
```
kvark: this method doesn't mutate the actual byte stream even on success?

torkleyy: It does, but through calling `self.deserialize_tuple` / `self.deserialize_struct`. I need to create a working copy for lookahead first, the bytes do get consumed later.

torkleyy:
Thank you for the review!
bors r+

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ron"
-version = "0.3.0"
+version = "0.4.0"
```
matklad: Would you like to make a release? I am hitting this exact issue right now :rofl:

torkleyy:
Sure!

torkleyy:
Done ;)

Issue #123, opened 2018-08-05
Externally, internally, adjacently and un-tagged enums


torkleyy:
All of those need to be properly tested, I don't think they are working currently. Behavior wrt. `Value` needs to be documented.

Also see: https://serde.rs/enum-representations.html

Boscop:
@torkleyy I just ran into this, a normal enum like this is serialized as `()`:
```
#[derive(Serialize, Deserialize)]
pub enum Foo {
	A,
	B,
	C,
}
Any idea when enums will be supported?

**Comment by kvark** [comment]
@Boscop enums in general are (of course!) supported.
I just tried to add your enum to the `encode.rs` example:
```diff
diff --git a/examples/encode.rs b/examples/encode.rs
index 3d3dc25..bf3b1e1 100644
--- a/examples/encode.rs
+++ b/examples/encode.rs
@@ -10,6 +10,7 @@ struct Config {
     nested: Nested,
     var: Variant,
     array: Vec<()>,
+    foo: Foo,
 }
 
 #[derive(Serialize)]
@@ -26,6 +27,13 @@ struct Nested {
     b: char,
 }
 
+#[derive(Serialize)]
+enum Foo {
+    A,
+    B,
+    C,
+}
+
 fn main() {
     let data = Config {
         float: (2.18, -1.1),
@@ -37,6 +45,7 @@ fn main() {
         },
         var: Variant::A(!0, ""),
         array: vec![(); 3],
+        foo: Foo::C,
     };
 
     let pretty = PrettyConfig::new()
```
And it serializes just fine:
```
(
    float: (2.18,-1.1,),
    tuple: ((),false,),
    map: {1:'2',3:'5',0:'1',8:'1',},
    nested: (a:"Hello from \"RON\"",b:'b',),
    var: A(255,"",),
    array: [(),(),(),],
    foo: C,
)
```

This issue is about funky custom-configured enums that Serde supports, not regular stuff.

Boscop:
@kvark Hm, but for me it serialized it as `()`, any idea why?

kvark:
@Boscop you are likely talking about a different issue. Could you file one, provide the repro steps, and we'll go from there?

OvermindDL1:
Might be related to this, but I've tried each of the 3 enum types.  Basically I have an array of an enum, and I'm trying to get something like this:
```ron
[
[...]

Issue #122, opened 2018-08-05
Current issues with `Value`


torkleyy:
* [x] tuples / tuple structs not supported (#121)
* [ ] enums not supported
* [ ] struct names don't get stored
* [ ] units represented with struct names
* [ ] Make sure errors make sense, esp after a unit struct (e.g. `UnitStruct:` as input)
* [ ] figure out what to do about tuples becoming sequences

Boscop:
Can some of these boxes be ticket now?

kvark:
comment

wiiznokes:
Any plans on improving this or this is out of scope ?

juntyr:
Yes, many plans, little time (research + small child) :/ If you can help with any part, I'd really appreciate it :)

wiiznokes:
> Yes, many plans, little time (research + small child) :/ If you can help with any part, I'd really appreciate it :)

I'm interested in making the conversion "ron str" -> `ron::Value` lossless. Maybe i could try to implement it, but i think it would help if you can give some hint about the change needed in the Value type.

juntyr:
> > Yes, many plans, little time (research + small child) :/ If you can help with any part, I'd really appreciate it :)
> 
> I'm interested in making the conversion "ron str" -> `ron::Value` lossless. Maybe i could try to implement it, but i think it would help if you can give some hint about the change needed in the Value type.

Essentially the Value type would need to become an AST-like type for everything the current grammar can represent. Then ron string -> Value -> string would be lossless.

You can find the grammar here: https://github.com/ron-rs/ron/blob/master/docs/grammar.md (it even has a Value section).

wiiznokes:
Ok, i've got smtg like this:

```rs
pub enum Value {
    Bool(bool),
    Char(char),
    Map(Map),
    Struct(Option<String>, Map),
    Number(Number),
    Option(Option<Box<Value>>),
    String(String),
    Bytes(Vec<u8>),
    Seq(Vec<Value>),
    Tuple(Option<String>, Vec<Value>),
    Unit,
}
```

wiiznokes:
> we also need a named unit variant with a string field (e.g. for unit structs and unit variants)

I was thinking of using `Struct(None, empty::map)` for "()", and `Struct(Some("name"), empty::map)` for unit_struct with indent.
For enum_variant_unit, we can use `Tuple(Some("name"), empty::vec)`.

[...]

Issue #115, opened 2018-06-10
Struct flattening not working


ebkalderon:
I'm trying to use Serde's [struct flattening](https://serde.rs/attr-flatten.html) functionality with RON, using the Rust code below:

```rust
#[derive(Deserialize, Serialize)]
struct Main {
    #[serde(flatten)]
    required: Required,
    #[serde(flatten)]
    optional: Optional,

    some_other_field: u32,
}

#[derive(Deserialize, Serialize)]
struct Required {
    first: u32,
    second: u32,
}

#[derive(Deserialize, Serialize)]
struct Optional {
    third: Option<u32>,
}
```

This is the RON that I am attempting to deserialize:

```ron
Main(
    first: 1,
    second: 2,
    third: Some(3),
    some_other_field: 1337,
)
```

When I try to deserialize the target struct from RON, I get the following error:

```
thread 'example::test' panicked at 'called `Result::unwrap()` on an `Err` value: Parser(ExpectedMap, Position { col: 9, line: 2 })', libcore/result.rs:945:5        
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

On the other hand, deserializing from the following TOML string works as intended:

```toml
first = 1
second = 2
third = 3
some_other_field = 4
```

Any indication on what might be going wrong?

kvark:
Interesting. I haven't worked with struct flattening before. Wondering if we need any special handling for those.

ebkalderon:
@kvark It looks like there are two kinds of struct flattening available in Serde. There's flattening into a map, which acts as a "catch-all" for fields that aren't found in your struct, like this:

```rust
#[derive(Deserialize, Serialize)]
struct MyType {
    first: u32
    second: u32,
    #[serde(flatten)]
    everything_else: HashMap<String, Value>,
}
```

Then there's flattening into a struct, which operates in a similar manner and is demonstrated in the example code I showed in the first comment of the issue.

Not sure if this needs any special handling on RON side either, but I can't think of what else might be going wrong. It appears that `toml` handles it correctly, but I can't think of what they might be doing different.

torkleyy:
Blocked by [`serde-rs/serde#1346`](https://github.com/serde-rs/serde/issues/1346)

torkleyy:
Well, it seems we need some workaround since the serde issue won't be fixed.

skysch:
I've hit this problem, and found a workaround by using both a struct and map together.

I have an enum variant that looks like this:
```rust
[...]

PR #127, opened 2018-08-07
Check heterogeneous sequence doesn't become a list

torkleyy:


torkleyy:
[Spoiler alert]
It doesn't.

torkleyy:
I don't know how to solve this. Currently tuples and lists share one `Value` variant, because JSON and other formats don't have the distinction either. I could theoretically make it two, but there's no way of telling what it is when I'm deserializing - there's only `visit_seq`, which is for tuples and lists.

torkleyy:
I can't pass that information from the deserializer to the `Value` visitor..

torkleyy:
We need to consider making some contributions to serde and / or providing features that are not using serde. The current architecture design causes a couple of issues:

* this
* #115 
* #122
* #123

torkleyy:
Closing due to inactivity.

PR #148, opened 2019-02-28
Add integer check for deserializer

vessd:
Fix a problem with untagged enum deserialization

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -324,6 +324,28 @@ impl<'a> Bytes<'a> {
             .fold(0, |acc, _| acc + 1)
     }
 
+    pub fn next_bytes_is_float(&self) -> bool {
+        if let Some(byte) = self.peek() {
+            let skip = match byte {
+                b'+' | b'-' => 1,
+                _ => 0,
+            };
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -324,6 +324,37 @@ impl<'a> Bytes<'a> {
...
     }
 
+    pub fn next_bytes_is_float(&self) -> bool {
+        if let Some(byte) = self.peek() {
+            let mut skip = 0;
+            match byte {
+                b'+' | b'-' => skip = 1,
+                b'0' => {
+                    if self.bytes.get(1)
+                        .filter(|&&b| "box".contains(b as char))
```
kvark: why are we parsing "box"?

vessd: if a leading 0 is present, then we check if it is binary, octal or hexadecimal
`0b`, `0o` or `0x`

the parser provides such case, but I'm not sure that it is reachable right now
https://github.com/ron-rs/ron/blob/master/src/parse.rs#L467

vessd: I realized that this is not necessary

PR #150, opened 2019-03-05
don't insert new lines in empty arrays or maps

### LINKED ISSUE #147: Don't split empty sets/arrays into multiple lines
I noticed a lot of sections like this:
```ron
                            vertical_blurs: [
                            ],
                            horizontal_blurs: [
                            ],
                            scalings: [
                            ],
                            blits: [
                            ],
                            outputs: [
                            ],
```

I think it would be strictly an improvement if we detect that a set/map/array is empty and avoid inserting a new line in this case.

DivineGod:
This fixes #147

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -167,10 +167,8 @@ impl Serializer {
         if let Some((ref config, ref mut pretty)) = self.pretty {
             pretty.indent += 1;
             if pretty.indent < config.depth_limit {
-                let is_empty = match self.is_empty {
-                    Some(n) => n,
-                    None => false,
-                };
+                let is_empty = self.is_empty.unwrap_or(false);
+
                 if !is_empty {
                     self.output += &config.new_line;
                 }
@@ -190,10 +188,7 @@ impl Serializer {
     fn end_indent(&mut self) {
         if let Some((ref config, ref mut pretty)) = self.pretty {
             if pretty.indent < config.depth_limit {
-                let is_empty = match self.is_empty {
-                    Some(n) => n,
-                    None => false,
-                };
+                let is_empty = self.is_empty.unwrap_or(false);
 
                 if !is_empty {
                     self.output
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -18,6 +18,7 @@ where
         output: String::new(),
         pretty: None,
         struct_names: false,
+        is_empty: None,
     };
     value.serialize(&mut s)?;
     Ok(s.output)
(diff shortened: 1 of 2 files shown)


src/ser/mod.rs
```diff
@@ -179,11 +181,11 @@ impl Serializer {
 
     fn end_indent(&mut self) {
         if let Some((ref config, ref mut pretty)) = self.pretty {
+            pretty.indent -= 1;
             if pretty.indent < config.depth_limit {
                 self.output
                     .extend((1..pretty.indent).map(|_| config.indentor.as_str()));
```
kvark: isn't this changed now? I think you'd want to do `0..pretty.indent` if you subtract 1 earlier

DivineGod: Good point.

Testing it out just now, I think the mechanism is insufficient to handle empty sequences and maps because deeper nested items will still have a `pretty.indent > 0`

I moved it around, because the test failed with extra white-space.

kvark: But this would break the non-empty sequences, right?

DivineGod: Yeah, so I've moved it back in the latest iteration, added test data to handle deeper nested empty data sets.

kvark: I just didn't see the code update, hence the confusion :)

src/ser/mod.rs
```diff
@@ -163,7 +167,13 @@ impl Serializer {
         if let Some((ref config, ref mut pretty)) = self.pretty {
             pretty.indent += 1;
             if pretty.indent < config.depth_limit {
-                self.output += &config.new_line;
+                let is_empty = match self.is_empty {
```
kvark: could be `unwrap_or(false)`

src/ser/mod.rs
```diff
@@ -180,10 +190,19 @@ impl Serializer {
     fn end_indent(&mut self) {
         if let Some((ref config, ref mut pretty)) = self.pretty {
             if pretty.indent < config.depth_limit {
-                self.output
-                    .extend((1..pretty.indent).map(|_| config.indentor.as_str()));
+                let is_empty = match self.is_empty {
```
kvark: same here

kvark:
@torkleyy heads up in case you want to take a look

src/ser/mod.rs
DivineGod: Thanks, I'm still learning all the ways of Rust, `unwrap_or(false)` will make that a lot simpler.

kvark:
Thank you!
Bors r+

PR #157, opened 2019-04-05
Be smarter about integer deserialization

### LINKED ISSUE #100: deserialize_any treats integers as f64
The [current implementation](https://github.com/ron-rs/ron/blob/a5a4b2ee21f5be0b07c55844e2689ae8a521da3b/src/de/mod.rs#L124) of `deserialize_any` treats all numeric values as f64. 

This causes issues when using crates such as [serde-value](https://github.com/arcnmx/serde-value) as it will deserialize to the F64 variant preventing it from subsequently being able to be deserialized to an integer field, e.g.:

    struct Config {
        value: u32
    }

    fn value_deserialize_ron() {
        let serialized = r#"(
            value: 5
        )"#;

        let value = ::ron::de::from_str::<::serde_value::Value>(serialized).unwrap();

        let config = value.deserialize_into::<Config>().expect("failed to deserialize from value");
    }

This fails on the last line with the following error:

> failed to deserialize from value: InvalidType(Float(5.0), "u32")

Both serde yaml:

    fn value_deserialize_yaml() {
        let serialized = r#"
            value: 5
        "#;

        let value = ::serde_yaml::from_str::<::serde_value::Value>(serialized).unwrap();

        let config = value.deserialize_into::<Config>().expect("failed to deserialize from value");
    }

And serde json:

    fn value_deserialize_json() {
        let serialized = r#"{
            "value": 5
        }"#;

        let value = ::serde_json::from_str::<::serde_value::Value>(serialized).unwrap();

        let config = value.deserialize_into::<Config>().expect("failed to deserialize from value");
    }

Are able to handle this case correctly, so I believe the issue is with how Ron parses these numeric values.

### LINKED ISSUE #155: Deserializing large serialized `Value::Number` fails
*Expected behavior*: Deserialisation of any serialized struct should succeed.

*Actual behavior*: When deserializing a large serialized `Value::Number` an error is returned.

Example:
```rust
use ron::value::Value;
let test_var = Value::Number(Number::new(10000000000000000000.0f64));
let test_ser = ron::ser::to_string(&test_var).unwrap();
let test_deser = ron::de::from_str::<Value>(&test_ser);
assert!(test_deser.is_ok()); // <-- this fails
```
Here `test_deser` is an `Err<Parser(ExpectedInteger, Position { col: 1, line: 1 })>`.

Version:
`git rev-parse HEAD`: `6e9f7dd3080bdedb88ad89700da1aad499cb303f`

torkleyy:
Now, `deserialize_any` will try to find the best representation possible.

This is not a very pretty implementation, but it should work well.

Fixes #100 
Fixes #155

### DIFF
--- a/src/de/tests.rs
+++ b/src/de/tests.rs
@@ -305,3 +305,18 @@ fn test_numbers() {
         from_str("[1_234, 12_345, 1_2_3_4_5_6, 1_234_567, 5_55_55_5]"),
     );
 }
+
+fn de_any_number(s: &str) -> AnyNum {
+    let mut bytes = Bytes::new(s.as_bytes()).unwrap();
+
+    bytes.any_num().unwrap()
+}
+
+#[test]
+fn test_any_number_precision() {
+    assert_eq!(de_any_number("1"), AnyNum::U8(1));
+    assert_eq!(de_any_number("+1"), AnyNum::I8(1));
+    assert_eq!(de_any_number("-1"), AnyNum::I8(-1));
+    assert_eq!(de_any_number("-1.0"), AnyNum::F32(-1.0));
+    assert_eq!(de_any_number("0.3"), AnyNum::F64(0.3));
+}
--- a/src/de/tests.rs
+++ b/src/de/tests.rs
@@ -150,8 +150,6 @@ y: 2.0 // 2!
 }
 
 fn err<T>(kind: ParseError, line: usize, col: usize) -> Result<T> {
-    use crate::parse::Position;
-
     Err(Error::Parser(kind, Position { line, col }))
 }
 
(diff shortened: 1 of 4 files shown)


src/de/tests.rs
```diff
@@ -305,3 +305,18 @@ fn test_numbers() {
...
+
+    bytes.any_num().unwrap()
+}
+
+#[test]
+fn test_any_number_precision() {
+    assert_eq!(de_any_number("1"), AnyNum::U8(1));
+    assert_eq!(de_any_number("+1"), AnyNum::I8(1));
+    assert_eq!(de_any_number("-1"), AnyNum::I8(-1));
+    assert_eq!(de_any_number("-1.0"), AnyNum::F32(-1.0));
```
azriel91: Tangential: Rust also accepts `1.` and `-1.`, does the parser accept them?

tests/large_number.rs
```diff
@@ -0,0 +1,12 @@
+use ron::value::Number;
+
+#[test]
+fn test_large_number() {
+    use ron::value::Value;
+    let test_var = Value::Number(Number::new(10000000000000000000000.0f64));
+    let test_ser = ron::ser::to_string(&test_var).unwrap();
+    println!("test_ser: {}", test_ser);
```
azriel91: intentionally left in?

torkleyy:
Thanks for the review, addressed your comments.

src/de/tests.rs
torkleyy: Yes.

PR #161, opened 2019-04-18
Parse integers ourselves, fix MIN parsing

### LINKED ISSUE #158: Deserializing std::i*::MIN integer fails
RON can serialize `std::i32::MIN` but fails to deserialize it.
That's because it's trying [to parse](https://github.com/ron-rs/ron/blob/64ba4a541204defb8218e75b7bea7c3a7c771226/src/parse.rs#L526) `-2147483648i32` as `2147483648i32` [using i32::from_str_radix](https://github.com/ron-rs/ron/blob/64ba4a541204defb8218e75b7bea7c3a7c771226/src/parse.rs#L675) and then [negate it](https://github.com/ron-rs/ron/blob/64ba4a541204defb8218e75b7bea7c3a7c771226/src/parse.rs#L388). But `i32::from_str_radix` fails because `std::i32::MAX` is `2147483647`.
Haven't checked, but most probably there is same problem with `i8::MIN`, `i16::MIN`, `i64::MIN` and `isize::MIN`.

torkleyy:
Fixes #158

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -3,7 +3,6 @@ use std::{
     char::from_u32 as char_from_u32,
     fmt::{Display, Formatter, Result as FmtResult},
     ops::Neg,
-    result::Result as StdResult,
     str::{from_utf8, from_utf8_unchecked, FromStr},
 };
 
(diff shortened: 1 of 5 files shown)


src/parse.rs
```diff
@@ -84,6 +83,78 @@ impl<'a> Bytes<'a> {
...
+            return self.err(ParseError::ExpectedInteger);
+        }
+
+        let s = unsafe { from_utf8_unchecked(&self.bytes[0..num_bytes]) };
+
+        if s.as_bytes()[0] == b'_' {
+            return self.err(ParseError::UnderscoreAtBeginning);
+        }
+
+        fn calc_num<T: Num>(
```
kvark: can we just return an iterator of digits here?

torkleyy: Yes, we could, but then we'd have to duplicate the multiplication.

PR #163, opened 2019-05-10
Implement inf, -inf and NaN handling

### LINKED ISSUE #159: f64::INFINITY causes panic
# Problem summary

I need to be able to serialize and deserialize `f64::INFINITY` in my code.  I believe that this should be possible, but the following code causes a panic when run:

```rust
use ron::de;
use ron::ser;
use serde::*;
use std::cmp::{Eq, PartialEq};
use std::f64;

#[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
pub struct Test1 {
    field: f64,
}

impl PartialEq for Test1 {
    fn eq(&self, other: &Test1) -> bool {
        self.field == other.field
    }
}

impl Eq for Test1 {}

fn main() {
    let original = Test1 {
        field: f64::INFINITY,
    };

    let pretty: ser::PrettyConfig = Default::default();
    let serialized = ser::to_string_pretty(&original, pretty.clone()).unwrap();
    let duplicate: Test1 = de::from_str(&serialized).unwrap();
    assert_eq!(original, duplicate);
}
```
There is no panic for finite `f64` values.

The panic:
```
$ cargo run
   Compiling serde_float_issue v0.1.0 (/home/cfkaran2/Desktop/bug_reports/serde_float_issue)
    Finished dev [unoptimized + debuginfo] target(s) in 0.49s
     Running `target/debug/serde_float_issue`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Parser(ExpectedFloat, Position { col: 12, line: 2 })', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

# Meta information:

```$ uname -a ; echo ; lsb_release --all ; echo ; rustc -Vv ; echo ; cargo -Vv
Linux EMANE 4.15.0-47-generic #50-Ubuntu SMP Wed Mar 13 10:44:52 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic

rustc 1.34.0 (91856ed52 2019-04-10)
binary: rustc
commit-hash: 91856ed52c58aa5ba66a015354d1cc69e9779bdf
commit-date: 2019-04-10
host: x86_64-unknown-linux-gnu
release: 1.34.0
LLVM version: 8.0

cargo 1.34.0 (6789d8a0a 2019-04-01)
release: 1.34.0
commit-hash: 6789d8a0a54a96d95365c4e1fb01d47a5eed9937
commit-date: 2019-04-01
```

artemshein:
FIxes #159

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -364,6 +364,13 @@ impl<'a> Bytes<'a> {
     where
         T: FromStr,
     {
+        for literal in &["inf", "-inf", "NaN"] {
+            if self.consume_ident(literal) {
+                return FromStr::from_str(literal)
+                    .map_err(|_| unreachable!()); // must not fail
+            }
+        }
+
         let num_bytes = self.next_bytes_contained_in(FLOAT_CHARS);
 
         let s = unsafe { from_utf8_unchecked(&self.bytes[0..num_bytes]) };
--- a/src/value.rs
+++ b/src/value.rs
@@ -16,17 +16,11 @@ use crate::de::{Error as RonError, Result};
 
 /// A wrapper for `f64` which guarantees that the inner value
 /// is finite and thus implements `Eq`, `Hash` and `Ord`.
-#[derive(Copy, Clone, Debug, PartialOrd, PartialEq)]
+#[derive(Copy, Clone, Debug)]
 pub struct Number(f64);
 
 impl Number {
-    /// Panics if `v` is not a real number
-    /// (infinity, NaN, ..).
     pub fn new(v: f64) -> Self {
-        if !v.is_finite() {
-            panic!("Tried to create Number with a NaN / infinity");
-        }
-
         Number(v)
     }
 
@@ -36,6 +30,22 @@ impl Number {
     }
 }
 
+/// Partial equality comparison
+/// In order to be able to use `Number` as a mapping key, NaN floating values wrapped in `Number`
+/// are equals to each other. It is not the case for underlying `f64` values itself.
+impl PartialEq for Number {
+
+    fn eq(&self, other: &Self) -> bool {
+        if self.0.is_nan() && other.0.is_nan() {
+            return true;
+        }
+        return self.0 == other.0;
+    }
+}
+
+/// Equality comparison
+/// In order to be able to use `Number` as a mapping key, NaN floating values wrapped in `Number`
+/// are equals to each other. It is not the case for underlying `f64` values itself.
 impl Eq for Number {}
 
 impl Hash for Number {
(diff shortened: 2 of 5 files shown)


src/parse.rs
```diff
@@ -364,6 +364,17 @@ impl<'a> Bytes<'a> {
     where
         T: FromStr,
     {
+        if self.consume_ident("inf") {
```
kvark: sounds like these conditions could be simplified (with a loop or something)

src/value.rs
```diff
@@ -20,13 +20,7 @@ use crate::de::{Error as RonError, Result};
 pub struct Number(f64);
 
 impl Number {
-    /// Panics if `v` is not a real number
-    /// (infinity, NaN, ..).
     pub fn new(v: f64) -> Self {
-        if !v.is_finite() {
-            panic!("Tried to create Number with a NaN / infinity");
-        }
```
torkleyy: I don't think we can just remove this; we can only implement `Ord` because of this check. Can you please check what behavior this causes when sorting `Number`?

artemshein: We don't implement Ord, only PartialOrd:
```
#[derive(Copy, Clone, Debug, PartialOrd, PartialEq)]
pub struct Number(f64);
```
and it should be fine as:
```
impl PartialOrd<f64> for f64
```

artemshein: Yes, you're right, it will fail in case of NaN:
```
impl Ord for Number {
    fn cmp(&self, other: &Self) -> Ordering {
        self.partial_cmp(other).expect("Bug: Contract violation")
    }
}
```
Probably we should not provide Ord implementation for Number.

torkleyy: I don't think we can not provide the `Ord` implementation, because `Number` is a variant of `Value`, which in turn needs to implement `Ord` so we can use `Value` as key.

So, what we should do is just handle the `NaN` case and specify the ordering in the docs. For example, we might just always return `Less` for `NaN`.

artemshein: It means we also need to change Eq implementation and make two NaNs equals to each other. And also make Ord return Less if the right is Nan but the left is not, Equal is both are NaNs and Greater if the right is not NaN, but the left is. Correct?

artemshein: Implemented

src/parse.rs
artemshein: Done

kvark:
Please rebase instead of merge committing

torkleyy:
Um, bors?
bors ping

PR #171, opened 2019-06-06
Note that raw strings are not context-free

### LINKED ISSUE #167: raw strings in rust are context-sensitive
The grammar docs note:

> Note: Raw strings start with an r, followed by n # and a quotation mark ". They may contain any characters or escapes (except the end sequence). A raw string ends with a quotation mark ("), followed by n #.
>
> I don't know any sane way to write this out in EBNF, if you do, let me know.

It can't be written in EBNF because the syntax of EBNF is for expressing context-free grammars, while raw strings are context-sensitive:

https://github.com/rust-lang/rust/blob/master/src/grammar/raw-string-literal-ambiguity.md

(As a general rule, context-free languages can only compare two numbers destructively, so comparing 3 numbers doesn't work)

torkleyy:
Fixes #167

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -71,7 +71,10 @@ string_raw = ("r#", string_raw, "#") | "\"", { unicode_non_greedy }, "\"";
 r##"This is a "raw string". It can contain quotations or
 backslashes (\)!"##
   ```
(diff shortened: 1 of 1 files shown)


torkleyy:
bors ping

PR #172, opened 2019-06-06
Use indexmap to preserve order (optional)

### LINKED ISSUE #129: Feature request: preserve order in value
Currently, value is based on `BTree` map, which loses the original order of ron document. I think it makes sense to add an option to preserve order. I'd even say that if ron's primary purpose is to be a human-editable format, then order should be preserved by default.  https://crates.io/crates/indexmap might come in handy do actually implement this. 

Here's my specific use-case for ordering: I [use ron](https://github.com/matklad/libsyntax2/blob/de02d2891ec7f7b295e33887d49de954f677387a/crates/libsyntax2/src/grammar.ron) to generate the AST for Rust. As such, I group related things, like types or expressions, together, and ideally I'd love to see the same order in the generated code. Alas, as I use [a map](https://github.com/matklad/libsyntax2/blob/de02d2891ec7f7b295e33887d49de954f677387a/crates/libsyntax2/src/grammar.ron#L218-L235) for some part of my input, it doesn't work. I can swtich to a list of pairs for this case, but a mapping makes more sense.

torkleyy:
Introduces an optional `indexmap` feature which uses `IndexMap` in `Value::Map` to preserve the order of deserialized elements.

Fixes #129

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1,6 +1,7 @@
 //! Value module.
 
 use serde::{
+    Deserialize, Serialize,
     de::{
         DeserializeOwned, DeserializeSeed, Deserializer, Error as SerdeError, MapAccess, SeqAccess,
         Visitor,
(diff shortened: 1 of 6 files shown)


src/value.rs
```diff
@@ -5,16 +5,132 @@ use serde::{
...
+impl PartialOrd for Map {
+    fn partial_cmp(&self, other: &Map) -> Option<Ordering> {
+        self.iter().partial_cmp(other.iter())
+    }
+}
+
+#[cfg(not(feature = "indexmap"))]
+type MapInner = std::collections::BTreeMap<Value, Value>;
+#[cfg(feature = "indexmap")]
+type MapInner = indexmap::IndexMap<Value, Value>;
```
kvark: what happens if we just typedef `type Map = indexmap::IndexMap<..>` instead of wrapping it?

torkleyy: We'll end up with a non-additive feature. `BTreeMap` and `IndexMap` have different properties (traits, methods, etc.).

Also, APIs accepting `BTreeMap` would break once `indexmap` is enabled.

torkleyy:
> If I understand correctly, this would only preserve the order of Value types but not regular maps?

Exactly. If one wants to preserve the order elsewhere, they should use `IndexMap` there, too.

torkleyy:
Thanks for the review!
bors r+

PR #94, opened 2018-02-07
[WIP] Trim spaces/commas

kvark:
Follow-up to #93 that trims stuff

Issue #175, opened 2019-06-06
Current `Serializer` is hard to maintain


torkleyy:
## The problem

The current serializer provides all kind of config parameters, but as it is now, it became hard to maintain.
Many things are special cased, and code is duplicated across tuples / enums / sequences / maps.

Additionally, we're lacking tests for pretty serialization.

## How to fix this

We need to build an abstraction for comma separated values, which removes the duplication we have now.

Once that's done, special cases should be easier to identify.

torkleyy:
This is still a problem, although we have more tests now thanks to @MomoLangenstein which should make refactoring easier.

kvark:
weird that GH actions bot closes this, even thought there is a comment?

juntyr:
I think that #206 fixed the code duplication issue and #488 finally pushed the serialising code to 100% code coverage

PR #177, opened 2019-06-06
Don't use trailing commas when no newlines are used

torkleyy:
This partially addresses #94; I will cover the other part of it (spaces between elements) after the serializer refactor (#175).

Closes #94

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -176,9 +176,22 @@ impl Serializer {
         }
     }
 
+    fn indentation_enabled(&self) -> bool {
+        // TODO(torkleyy): consider changing this to <=
+        // TODO: (see https://github.com/ron-rs/ron/issues/174)
+        self.pretty.as_ref().map(|(config, pretty)| pretty.indent < config.depth_limit)
+            .unwrap_or(false)
+    }
+
+    fn pop_if_comma(&mut self) {
+        if self.output.as_bytes().last() == Some(&b',') {
+            self.output.pop();
+        }
+    }
+
     fn indent(&mut self) {
         if let Some((ref config, ref pretty)) = self.pretty {
-            if pretty.indent < config.depth_limit {
+            if self.indentation_enabled() {
                 self.output
                     .extend((0..pretty.indent).map(|_| config.indentor.as_str()));
             }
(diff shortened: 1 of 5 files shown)


src/ser/mod.rs
```diff
@@ -176,16 +169,34 @@ impl Serializer {
...
                 self.output
                     .extend((0..pretty.indent).map(|_| config.indentor.as_str()));
             }
         }
     }
 
     fn end_indent(&mut self) {
+        // Remove trailing comma if no indentation is used
+        if !self.indentation_enabled() {
+            self.pop_if_comma();
```
kvark: could there be a better way to check this? i.e. pop if there was at least one item in the current block, instead of checking the output stream

src/ser/mod.rs
```diff
@@ -635,15 +649,15 @@ impl<'a> ser::SerializeStruct for &'a mut Serializer {
         self.output += key;
         self.output += ":";
 
-        if self.is_pretty() {
+        if self.indentation_enabled() {
```
kvark: I don't think this check is correct

src/ser/mod.rs
```diff
@@ -176,16 +169,34 @@ impl Serializer {
...
+    fn pop_if_comma(&mut self) {
+        if self.output.as_bytes().last() == Some(&b',') {
+            self.output.pop();
+        }
+    }
+
     fn indent(&mut self) {
         if let Some((ref config, ref pretty)) = self.pretty {
-            if pretty.indent < config.depth_limit {
+            if self.indentation_enabled() {
```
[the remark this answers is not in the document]
torkleyy: Yeah, but I'm not quite sure how to fix it. Maybe this can be fixed together with #175.

src/ser/mod.rs
torkleyy: Yes, let's do that.

src/ser/mod.rs
torkleyy: `self.indentation_enabled()` and `self.is_pretty()` are equivalent, the former is just clearer so I removed `self.is_pretty`.

Ah, and it shouldn't check for indentation here! So this means there was a mistake before, which is that we should check if `self.pretty` is some, not if indentation is enabled (so it turns out I was confused by the naming myself...).

torkleyy: Hmm, turns out we simply don't use any spaces right now if the depth limit was reached.

torkleyy:
Okay, so I don't think I can fix any more issues in this PR; the problem with inserting spaces is that my (bad) `pop_if_comma` implementation doesn't work, and the problem with fixing that is that I essentially need to rewrite the `Serializer` (see #175).
Given that the added code is not great, this PR probably isn't an improvement.

kvark:
Closing it as stale for now, please feel free to revive!

PR #182, opened 2019-07-13
Allow pretty ser to work with implicit-some extension

cmaher:
Additionally configure newline for tests to work on windows.

I was converting my yaml files to ron and needed this, so I figured I'd submit a PR.  Really enjoying RON as a data language, btw!

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -196,10 +196,6 @@ fn default_enumerate_arrays() -> bool {
     false
 }
 
-fn default_extensions() -> Extensions {
-    Extensions::empty()
-}
-
 impl Default for PrettyConfig {
     fn default() -> Self {
         PrettyConfig {
@@ -208,7 +204,7 @@ impl Default for PrettyConfig {
             indentor: default_indentor(),
             separate_tuple_members: default_separate_tuple_members(),
             enumerate_arrays: default_enumerate_arrays(),
-            extensions: default_extensions(),
+            extensions: Extensions::default(),
             _future_proof: (),
         }
     }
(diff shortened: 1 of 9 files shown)


src/ser/mod.rs
```diff
@@ -112,6 +96,8 @@ pub struct PrettyConfig {
     /// Enumerate array items in comments
     #[serde(default = "default_enumerate_arrays")]
     pub enumerate_arrays: bool,
+    /// Enable implicit_some extension
+    pub implicit_some: bool,
```
kvark: We would potentially want other features there as well. Let's make it a bitflags field to account for future needs?
We could move the `struct_names` in these flags too, sometimes later.

cmaher: Sure. I'm going to be busy for the next week, but I'll try to get to it after that.

cmaher:
Sorry for the delay-- I factored out the Extensions from `parse`, and adding it as a field on `PrettyConfig` (replacing the `implicit_some` in the initial commit). It's a bit awkward to not have it handle the struct_names, but I didn't see a clean way of handling that without breaking the existing API, which I didn't want to do here.

src/ser/mod.rs
```diff
@@ -207,6 +208,7 @@ impl Default for PrettyConfig {
             indentor: default_indentor(),
             separate_tuple_members: default_separate_tuple_members(),
             enumerate_arrays: default_enumerate_arrays(),
+            extensions: default_extensions(),
```
kvark: should probably just use `Extensions::default()` here

cmaher: ah, good point. will fix that right up.

PR #184, opened 2019-07-14
Added to_writer helper function

Mathspy:
Adding function parallel to `from_reader`

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -2,15 +2,27 @@ use serde::{ser, Deserialize, Serialize};
 use std::{
     error::Error as StdError,
     fmt::{Display, Formatter, Result as FmtResult, Write},
+    io,
 };
 
 mod value;
+
+/// A convenience function for writing data to a writer
+pub fn to_writer<W, T: ?Sized>(writer: W, value: &T) -> Result<()>
+where
+    W: io::Write,
+    T: Serialize,
+{
+    { writer }
+        .write_all(to_string(value)?.as_bytes())
+        .map_err(|err| err.into())
+}
 
 /// Serializes `value` and returns it as string.
 ///
 /// This function does not generate any newlines or nice formatting;
 /// if you want that, you can use `to_string_pretty` instead.
-pub fn to_string<T>(value: &T) -> Result<String>
+pub fn to_string<T: ?Sized>(value: &T) -> Result<String>
 where
     T: Serialize,
 {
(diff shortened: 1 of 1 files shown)


src/ser/mod.rs
```diff
@@ -2,15 +2,27 @@ use serde::{ser, Deserialize, Serialize};
...
 
 mod value;
 
+/// A convenience function for writing data to a writer
+pub fn to_writer<W, T: ?Sized>(writer: W, value: &T) -> Result<()>
+where
+    W: io::Write,
+    T: Serialize,
+{
+    { writer }
```
kvark: why are "{" and "}" needed here?

src/ser/mod.rs
```diff
@@ -51,6 +63,8 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// Serialization error.
 #[derive(Clone, Debug, PartialEq)]
 pub enum Error {
+    /// io error from std::io
+    IoError(String),
```
kvark: this is changing the public API - a breaking change

Mathspy: Yep, I did it like this because that's how it's done in `de` but I don't mind doing it some other way

src/ser/mod.rs
Mathspy: So basically this creates a temporary variable binding (which are always `mut`) and then we chain the `write_all` (which requires `mut`) into that, it's a way to "dodge" the fact we don't ask for a `mut` writer as a parameter for this function.

The reason the function signature itself doesn't ask for a `mut` is because this is the same function signature other `serde` libraries uses:
[serde_json's to_writer](https://docs.serde.rs/serde_json/fn.to_writer.html)
[serde_yaml's to writer](https://docs.serde.rs/serde_yaml/fn.to_writer.html)
[Bincode's seralize_into](https://docs.rs/bincode/1.1.4/bincode/fn.serialize_into.html)

kvark: I see. I don't think `mut` in a function parameter is a part of the signature. It's purely an implementation detail that we shouldn't be shy of using, if it's relevant.

Mathspy: I don't mind changing it to `mut`, gonna amend that now

kvark:
@Mathspy what's the status of this PR?

src/ser/mod.rs
SirWindfield: Is it a breaking change because people matching on `Error` would have to match additionally on `IoError`? 🤔

kvark: yes, adding a variant to an exhaustive public enum is always a breaking change

Mathspy:
Oh yeah sorry! I left this PR behind for so long, I haven’t used RON in a while and I don’t really think the breaking changes in this PR were my call to make, if someone else wants the `to_writer` for RON please feel free to build on top of this, sorry for wasting everyone’s time! ><
And thanks for the continued awesome work all around (especially on the GPU side ❤️)

PR #190, opened 2019-10-09
Cargo version up Added Ryu crate to handle f32 and f64 serializing.

Hraesvel:
Chose to version up the crate to 0.6 since changes to serialize_f32 and serialize_f64 were made using Ryu rather than to_string() method.

serde_json uses Ryu, to stringify float values just an FYI.

The reasoning behind this change because to_string will strip the decimals from the value if given a whole value.
example being 1.0f32.to_string() will become "1" the correct result should have been "1.0"

This leads to an issue when deserializing the data back to its original type.

Additionally added 2 new `to_string` methods that includes the struct name.

other changes are result of Rustfmt

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ron"
-version = "0.6.0"
+version = "0.5.3"
 license = "MIT/Apache-2.0"
 keywords = ["parser", "serde", "serialization"]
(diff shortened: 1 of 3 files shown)


kvark:
@torkleyy please have a look, esp w.r.t. this being a non-breaking change.

Issue #192, opened 2019-10-13
Testing for valid ron files


fkarg:
It would be pretty awesome if you could provide an example as to how to test the validity of ron files with `cargo test` tests. With that, at least the 'tested' files can be guaranteed to be valid.

I imagine it to look something akin to this:

```rust
#[derive(Default)]
struct Point {
    x: f32,
    y: f32,
}


#[cfg(test)]
mod tests {
    #[test]
    fn load_ron_point() {

        let point: Option<Point> = load::<Point>("path/to/file.ron");
        assert_eq!(point.is_some(), true);

    }
}
```

And the test failing if for some reason the ron-file from the given path is no syntactically correct Point-type ron file.

kvark:
Isn't this covered by [decode_file](https://github.com/ron-rs/ron/blob/master/examples/decode_file.rs) example?

PR #197, opened 2019-11-03
Testing for RON files

peter-scholtens:
As a solution to the proposed test as mentioned in #192 I created a two functions: read_original()  and test_sequence(). Using an assert equal macro various files can be tested. Two examples are given: the first one fails as the mapping sequence changes, the secvond one is correct, as the map only contains one item.


tests/preserve_sequence.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+struct Nested {
+    a: String,
+    b: char,
+}
+
+fn read_original(file_name: &str) -> (String) {
+    let input_path = format!("{}/tests/{}", env!("CARGO_MANIFEST_DIR"), file_name);
```
kvark: Why don't we just include the original with `include_str!` or something?

tests/preserve_sequence.rs
```diff
@@ -0,0 +1,59 @@
...
+    fs::read_to_string(input_path)
+        .expect("Something went wrong reading the file")
+}
+
+fn test_sequence(file_name: &str) -> (String) {
```
kvark: perhaps this should be called `test_roundtrip` instead?

tests/preserve_sequence.rs
peter-scholtens: Didn't know it exists. However, if I try, the compiler keeps complaining "error: argument must be a string literal", unless I copy the string into the function. But then the flexibility of calling various files is gone.

tests/preserve_sequence.rs
peter-scholtens: Good suggestion. I though about several names (reciprocal, symmetric etc) but couldn't find a brief and correct description. See 1272b61ce01ae321d3e0050bdf6b6d87b8ea61d1

peter-scholtens: Forgot to update calling functions, see update in e3ada7f0c2876a598b17888cab2f9e445b3c9632 Now the first example fails as the mapping order is modified, while the second example proofs the test is okay.

tests/preserve_sequence.rs
kvark: What I was thinking:
```rust
fn read_original(source: &str) -> String {...}

fn test_xxx() {
  let file1 = include_str!("my_file1.ron");
  assert_eq!(
    read_original(file1),
    make_roundtrip(file1),
  );  
}
```

peter-scholtens: Ah, okay. Then the file is only read once, and processed twice. A much nicer way indeed. See 0fa9da321841aa6dea640007007348000a67f9a0

peter-scholtens:
Like this ? c0d52d20cf1d2f65e29af51f797a5bf347ba317a

peter-scholtens:
It should fail testing on the first example1 as the sequence is not preserved. Or should this be listed as a separate bug? The second test should pass succesfully.

kvark:
thank you!
bors r+

PR #209, opened 2020-04-12
Add positional information to error messages

halvnykterist:


halvnykterist:
This seems to cause a stack overflow under some circumstances so that's probably bad

src/de/mod.rs
```diff
@@ -126,21 +133,27 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         V: Visitor<'de>,
     {
         if self.bytes.consume_ident("true") {
-            return visitor.visit_bool(true);
+            return visitor.visit_bool(true).map_err(|e| self.fix_position(e));
```
kvark, line 136: why all these cases just don't return positional errors to begin with?

halvnykterist: If I understand correctly, the errors are generated by `fn custom<T: fmt::Display>(msg: T) -> Self` on `Error` in `error.rs`.  In serde's json deserializer ( https://github.com/serde-rs/json/blob/master/src/error.rs#L380 ) they've already got the line number and column built into the message, somehow, but I have a hard time following the path.

src/de/mod.rs
```diff
@@ -39,6 +39,13 @@ impl<'de> Deserializer<'de> {
     pub fn remainder(&self) -> Cow<'_, str> {
         String::from_utf8_lossy(&self.bytes.bytes())
     }
+
+    fn fix_position(&self, e: Error) -> Error {
```
kvark, line 43: Ok, let's just make it so the input error type and the output error type are *different*. This will still require us to call it, but there will be no way to miss a fix step.

halvnykterist: That sounds smart, I'll give that a try

Issue #47, opened 2017-08-10
[RFC] Separate number types


kvark:
See https://www.reddit.com/r/rust/comments/6sm6i2/rusty_object_notation/dlei3gp/

Basically, Rust clearly differentiates between integer types and floating points - there is no ambiguity. No reason for Ron to mix those types either.

Aceeri:
How about adding in typed numbers like in rust: `32u32`, `5.0f64`, etc. ?

kvark:
@Aceeri I think it would be an overkill. Loading everything as u64 and f64 should be good enough. After all, we are talking about a text format here, not the crazily optimized binary representation.

mrZalli:
What if you loaded integers as BigInts from [num](https://docs.rs/num/0.1.40/num/) crate since they can handle arbitrarily large numbers and then convert to the target integer or other type that can be converted from it. This would be great if [128-bit support](https://github.com/rust-lang/rust/issues/35118) ever comes to stable, or we want to use it in nightly, and we want to represent 128-bit number literals in ron text, instead of bypassing the 64-bit limitation with something like using a two 64-bit integer tuple or a quoted string representation, or you know if we wanted to store BigInts.

kvark:
@mrZalli hmm, wouldn't it be up to `BigInt` implementation to figure out how it's going to be serialized/deserialized?
128 bit would be nice but not essential, strictly speaking. Numbers as big are not readable anyway, so you might as well pass them as a hex string.

torkleyy:
Related: https://github.com/ron-rs/ron/issues/100

kvark:
@torkleyy 
> I don't think we need to be super concise here, so 5 as float is fine for me.

Do you find it better than just "5.0"? And if so, why?

torkleyy:
Not better, but also acceptable.

cart:
I have a use case that makes the `5` vs `5.0` distinction important. My rust type is a dynamic collection that can contain either floats or ints. Round tripping a value of `5.0` with ron results in an integer output of `5` when its source was a float.

In the short term I have forked ron to support this use case, but I'd love to get this upstreamed if possible:
https://github.com/ron-rs/ron/pull/237

SirWindfield:
Again, not sure about `serde` internals here but given the following struct:
```rust
#[derive(Deserialize)]
struct Config {
	float: f32,
    integer: u8,
}
```
and this config:
```ron
Config(
    float: 5.0,
    integer: 255
)
```

[...]

PR #210, opened 2020-04-18
Integer support

elrnv:
Roughly following the discussion from https://github.com/ron-rs/ron/issues/47 and related issues and PRs, this PR extends the Number type (and so indirectly the Value type) to support integers. I hope I didn't grossly misunderstand the discussion there, but I am keen to make adjustments here to make the project better.

~~I tried to limit the impact to the API to a single breaking change by extending the already opaque `Number` type rather than the `Value` enum which exposes its structure. The breaking change is that the `get` method in `Number` is replaced by `as_f64` and `as_i64` for getting the appropriate type.~~
Since `Integer` and `Float` types are distinct, they will not be ordered by value (as explained below). This is a breaking change in itself.

To maintain the traits on `Number`, `Integer`s are simply ordered before `Float`s. This makes float and integer not comparable in value (e.g. `Number::Integer(2)` < `Number::Float(Float(1.0))`) because I didn't see any necessity to do so, but I'm open to suggestions.

The tests are adjusted to verify that both integers and floats are working.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -131,14 +131,22 @@ type MapInner = std::collections::BTreeMap<Value, Value>;
 #[cfg(feature = "indexmap")]
 type MapInner = indexmap::IndexMap<Value, Value>;
 
-/// A wrapper for `f64` which guarantees that the inner value
+/// A wrapper for a number, which can be either `f64` or `i64`.
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd, Eq, Hash, Ord)]
+pub enum Number {
+    Integer(i64),
+    Float(Float),
+}
+
+/// A wrapper for `f64`, which guarantees that the inner value
 /// is finite and thus implements `Eq`, `Hash` and `Ord`.
 #[derive(Copy, Clone, Debug)]
-pub struct Number(f64);
+pub struct Float(f64);
 
-impl Number {
+impl Float {
+    /// Construct a new `Float`.
     pub fn new(v: f64) -> Self {
-        Number(v)
+        Float(v)
     }
 
     /// Returns the wrapped float.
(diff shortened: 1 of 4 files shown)


src/value.rs
```diff
@@ -147,11 +155,94 @@ impl Number {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use ron::value::Number;
+    /// let i = Number::new(5);
+    /// let f = Number::new(2.0);
+    /// assert_eq!(i.get(), 5.0);
+    /// assert_eq!(f.get(), 2.0);
+    /// ```
+    pub fn get(self) -> f64 {
```
kvark, line 176: should be called `into_f64`

kvark:
@torkleyy would you want to review? I figured you know the value semantics better :)

src/value.rs
elrnv: Should we have a deprectaed `get` that is an alias for `into_f64` for backwards compatibility?

kvark: I thought your PR is API breaking anyway, isn't it?

elrnv: You're right. sorry, since floats are strictly not equal to integers of the same value. Do you think it's worthwhile to try to preserve this equality at all for compatibility or otherwise?

elrnv:
The rebase is done. Is there anything else needed here from me?

kvark:
@elrnv no, thank you!
Let's ping @torkleyy again and give them a few days, then proceed.

kvark:
Sorry, this needs to be rebased again.
Otherwise, we are all good to go.

elrnv:
@kvark No problemo. Should good now.
FYI I added a couple of `From` impls for the `Number` type that do proper (lossless) conversion from `u64` and `i32`. Previously I just had u64 truncating to i64.

kvark:
@CAD97 looks like GHA isn't configured correctly. It doesn't even run on this PR.

kvark:
@elrnv thank you! Unfortunately we can't proceed until we get CI looking at this.

kvark:
Thank you! I think the CI fix need to go separately first.

elrnv:
I thought there was an option to run GHA on forks, but I can't find it anymore. I was just trying to see if this would work... :/

CAD97:
@kvark I set it up the same way that Travis was, to only run on pushes to master/staging/trying. We can set it up to run on PRs as well, that's just another entry to the `on` key.

kvark:
@CAD97 travis was definitely checking other forks PRs

CAD97:
#227 bumps up the CI to always run on any PR or any push. Bors would have (of course) still tested on CI.

kvark:
Can we rebase and squash please?

CAD97:
.... I have no idea why GHA aren't running on this PR now. I can only suggest a `bors try` or recreating the PR at this point.

kvark:
I'll try to close and re-open it.

kvark:
Doesn't seem to help :/

kvark:
Maybe related to:
> 
Update - We are continuing to work on a fix for the elevated error rates.
May 5, 01:34 UTC
Update - We have identified the issue causing elevated error rates and are working on a fix.
May 5, 01:09 UTC
Investigating - We are investigating elevated errors with GitHub.
May 5, 00:45 UTC

CAD97:
CI has run now, anyway.

Issue #211, opened 2020-04-18
Clippy + rustfmt project


elrnv:
Should this project run through clippy and rustfmt checks in CI.

Given that there is an explicit rustfmt.toml it is surprising that the diff is currently so large when running rustfmt on the project.

elrnv:
To add, I would be happy to submit a PR for this.

kvark:
I think we should take this opportunity and switch to rustfmt stable, removing all the options that are nightly-only.
Clippy would be great! but perhaps not if it's going to block the PRs because of warnings.

elrnv:
This is resolved by #215 and #214.

PR #214, opened 2020-04-21
Rustfmt stable

elrnv:
This PR:
1. Updates rustfmt.toml with stable features only (all nightly-only features removed).
2. Adds a `cargo fmt -- --check` to ,travis.yml

PR #215, opened 2020-04-21
Add clippy to CI + address clippy problems

elrnv:
This commit
1. Adds `cargo clippy` to .travis.yml that is allowed to fail to reduce friction for incoming PRs.
2. Addresses clippy concerns and clarifies places where they should be ignored.

Addresses half of #211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,4 @@ cache: cargo
 script:
   - cargo test
   - cargo test --all-features
+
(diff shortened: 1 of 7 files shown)


elrnv:
I think I had a `rustup component add clippy` missing here and same goes for #214. I updated both, hopefully this will satisfy the CI.

kvark:
Please rebase!

elrnv:
All done!

PR #219, opened 2020-04-30
Support i128 and u128

e00E:
This PR adds support for 128 bit integers.

RON does not support 128 bit integers. This is undocumented. The Readme wrongly claims that the full Serde data model is supported.

### DIFF
--- a/src/de/id.rs
+++ b/src/de/id.rs
@@ -71,6 +71,13 @@ impl<'a, 'b: 'a, 'c> de::Deserializer<'b> for &'c mut IdDeserializer<'a, 'b> {
         unimplemented!("IdDeserializer may only be used for identifiers")
     }
 
+    fn deserialize_i128<V>(self, _: V) -> Result<V::Value>
+    where
+        V: Visitor<'b>,
+    {
+        unimplemented!("IdDeserializer may only be used for identifiers")
+    }
+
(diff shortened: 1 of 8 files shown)


e00E:
We might want to use https://docs.serde.rs/serde/macro.serde_if_integer128.html Thoughts? 

> Data formats that wish to support Rust compiler versions older than 1.26 (or targets that lack 128-bit integers) may place the i128 / u128 methods of their Serializer and Deserializer behind this macro.
>
> Data formats that require a minimum Rust compiler version of at least 1.26, or do not target platforms that lack 128-bit integers, do not need to bother with this macro and may assume support for 128-bit integers.

bh2smith:
I just want to say that this would be a very good addition to the overall ecosystem and that it will become immediately valuable within our project;

https://github.com/gnosis/dex-services/pull/822

Without it, we fear that we may have to implement our own serializations 🤮 !


really nice clean fix by the way. Looks like these changes fit very naturally in here.

bh2smith:
It appears these is another place that needs `u128` support? I'm still getting an error when calling `ron::de::from_bytes`. Have not yet found the culprit.

kvark:
@bh2smith are you able to use the git master for now? Or do you need a release soon?

PR #224, opened 2020-05-02
unbreak bors

CAD97:
Turns out it learns about the tests with these names. [Actually verified this works on my fork.](https://github.com/CAD97/ron/pull/1)


kvark:
Thank you!

PR #225, opened 2020-05-02
Fix #174

### LINKED ISSUE #174: Depth limit seems to be off by one
Currently, a depth limit of `1` means that nothing gets indented (same for `0`). That seems wrong, as the limit should be the (inclusive) maximum depth.

This can be fixed by changing the `pretty.indent < config.ident` to a `<=` check.

@kvark Do you agree this is a bug?

CAD97:
Disclaimer: untested; just applies the patch recommended by @torkleyy in #174.

Closes #174.

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -224,7 +224,7 @@ impl Serializer {
 
     fn is_pretty(&self) -> bool {
         match self.pretty {
-            Some((ref config, ref pretty)) => pretty.indent < config.depth_limit,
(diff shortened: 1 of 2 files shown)


CAD97:
Looks like there is a test of the current behavior. I don't actually have this branch locally (and am working on drafting a rewrite version anyway) so I'll let you grab that test.

kvark:
Well, since you made a PR first, you ought to finish it by correcting the tests expectations :)

CAD97:
Oh, alright, I'll pull out my `git stash` and fix this 😄

kvark:
bors try

PR #226, opened 2020-05-02
Update serde version to one that supports 128-bit integers

kvark:
r? @CAD97


CAD97:
Looks fine to me

kvark:
bors r=CAD97

CAD97:
That's probably because of mixed spaces/tabs in the toml, though why that would be problematic I don't know

kvark:
my web-based commit was utter garbage 😅 
bors r=CAD97

PR #232, opened 2020-05-17
Enable external crates to link to the doc

a1phyr:
Enable external crates to reference `ron`'s types in their documentation.

Note that it should be updated with each new version.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,7 +57,7 @@ Serializing / Deserializing is as simple as calling `to_string` / `from_str`.
 
 !*/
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -57,6 +57,8 @@ Serializing / Deserializing is as simple as calling `to_string` / `from_str`.
 
 !*/
 
+#![doc(html_root_url = "https://docs.rs/ron/5.1")]
```
kvark, line 60: Please put a comment in `Cargo.toml` saying that this line needs to be adjusted accordingly

kvark, line 60: the version should be "0.5.1", not "5.1"

PR #237, opened 2020-05-23
Optionally always include float decimals

cart:
This adds the option to always include the decimal point in floats.

Currently ron always serializes `5.0` as `5`. I have a use case that makes the `5` vs `5.0` distinction important. My rust type is a dynamic collection that can contain either floats or ints. Round tripping a value of `5.0` with ron results in an integer output of `5` when its source was a float.

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -76,6 +76,9 @@ pub struct PrettyConfig {
     /// Enumerate array items in comments
     #[serde(default = "default_enumerate_arrays")]
     pub enumerate_arrays: bool,
+    /// Always include the decimal in floats
+    #[serde(default = "default_decimal_floats")]
+    pub decimal_floats: bool,
     /// Enable extensions. Only configures 'implicit_some' for now.
     pub extensions: Extensions,
     /// Private field to ensure adding a field is non-breaking.
@@ -141,6 +144,17 @@ impl PrettyConfig {
         self
     }
 
+    /// Configures whether floats should always include a decimal.
+    /// When false `1.0` will serialize as `1`
+    /// When true `1.0` will serialize as `1.0`
+    ///
+    /// Default: `false`
+    pub fn with_decimal_floats(mut self, decimal_floats: bool) -> Self {
+        self.decimal_floats = decimal_floats;
+
+        self
+    }
+
     /// Configures extensions
     ///
     /// Default: Extensions::empty()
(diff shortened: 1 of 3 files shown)


cart:
Also, I'd love to use rust's fmt syntax to tack on the `.0`, but I can't find a way to use the minimum amount of precision required. Ex: `write!(w, "{:.5}", 1.0)` will always include the decimal, but it will always use a precision of exactly 5 decimal places: `1.00000`. I don't think rounding here is desirable.

src/ser/mod.rs
```diff
@@ -363,11 +388,17 @@ impl<'a, W: io::Write> ser::Serializer for &'a mut Serializer<W> {
 
     fn serialize_f32(self, v: f32) -> Result<()> {
         write!(self.output, "{}", v)?;
+        if self.decimal_floats() && v - v.floor() == 0.0 {
```
kvark, line 391: why isn't this written as `v == v.floor()`?

tests/floats.rs
```diff
@@ -1,8 +1,30 @@
...
 fn test_inf_and_nan() {
     assert_eq!(from_str("inf"), Ok(std::f64::INFINITY));
     assert_eq!(from_str("-inf"), Ok(std::f64::NEG_INFINITY));
     assert_eq!(from_str::<f64>("NaN").map(|n| n.is_nan()), Ok(true))
 }
+
+#[test]
+fn decimal_floats() {
+    let pretty = PrettyConfig::new().with_decimal_floats(false);
+    let without_decimal = to_string_pretty(&1.0, pretty).expect("Serialization failed");
```
kvark, line 16: Do we need those `expect` strings here? For tests, I think `unwrap` is perfectly fine.

src/ser/mod.rs
cart: No reason. Your representation is clearly better :smile:

tests/floats.rs
cart: I was following the pattern in the "preserve_sequence" tests, but I'm happy to swap these out. "preserve_sequence" appears to be the only existing test that uses expect.

kvark:
Note: clippy isn't happy with float comparisons (red on CI)

cart:
i opted for the clippy recommended solution: ``` (v - v.floor()).abs() < f32::EPSILON```

cart:
oops :)

cart:
It looks like EPSILON associated constants aren't in 1.34. Is backward compat desirable? I can either inline the epsilon values or someone will need to bump that version to 1.43. What should I do here?

kvark:
Let's not raise the Rust version just yet. Debian and other distros are typically much behind the top of the tree.

PR #244, opened 2020-05-29
Use a simpler implementation for the `Compound` type

Plecra:
The code for it was ripped pretty verbatim from `serde-json`, even though we don't actually need an enum. I switched to a struct declaration while I had the repo open, which removes many cases of destructuring and makes the code simpler to follow.

This PR is built on #243, so that should probably be merged first


Plecra:
I imagine they'll ultimately be optimised identically. Using the enum made the syntax quite a lot noisier without any good reason.

```rust

Plecra:
I believe that did the trick?

Issue #217, opened 2020-04-26
Nested untagged and externally tagged enums fail to deserialize


caelunshun:
The following example deserializes an untagged enum with two variants using `ron`, `serde_json`, and `serde_yaml`. The latter two libraries output the expected result, but RON fails with a "data did not match any variant of untagged enum Either" error.

```rust
use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct Root {
    value: Either,
}

#[derive(Debug, Deserialize)]
#[serde(untagged)]
enum Either {
    One(One),
    Two(Two),
}

#[derive(Debug, Deserialize)]
enum One {
    OneA,
    OneB,
    OneC,
}

#[derive(Debug, Deserialize)]
enum Two {
    TwoA,
    TwoB,
    TwoC,
}

const RON_SRC: &str = "OneA";
const JSON_SRC: &str = "{ \"value\": \"OneA\" }";
const YAML_SRC: &str = "value: OneA";

fn main() {
    println!("{:?}", ron::de::from_str::<Either>(RON_SRC)); // Err(Message("data did not match any variant of untagged enum Either"))
    println!("{:?}", ron::de::from_str::<One>(RON_SRC)); // Ok(OneA)
    println!("{:?}", serde_json::from_str::<Root>(JSON_SRC)); // Ok(Root { value: One(OneA) })
    println!("{:?}", serde_yaml::from_str::<Root>(YAML_SRC)); // Ok(Root { value: One(OneA) })
}
```

caelunshun:
I've tracked this down to relate to the implementation of `handle_any_struct`:

```rust
    /// Called from `deserialze_any` when a struct was detected. Decides if
    /// there is a unit, tuple or usual struct and deserializes it
    /// accordingly.
    ///
    /// This method assumes there is no identifier left.
    fn handle_any_struct<V>(&mut self, visitor: V) -> Result<V::Value>
    where
        V: Visitor<'de>,
    {
        // Create a working copy
        let mut bytes = self.bytes;

        match bytes.consume("(") {
            true => {
                bytes.skip_ws()?;

                match bytes.check_tuple_struct()? {
                    // first argument is technically incorrect, but ignored anyway
                    true => self.deserialize_tuple(0, visitor),
                    // first two arguments are technically incorrect, but ignored anyway
                    false => self.deserialize_struct("", &[], visitor),
                }
            }
            false => visitor.visit_unit(),
        }
    }
```

[...]

PR #253, opened 2020-07-06
Make untagged enums work

joonazan:
When deserializing an untagged enum, serde uses deserialize_any. There is ambiguity, as we cannot know if a struct or an enum is expected. That ambiguity is resolved by serde if we just give it a JSON-like data structure.

When we encounter an lone identifier, it becomes a string. An identifier with parens after it becomes a singleton map.

Solves #217 

## Newtype pain

`Id(Val)` would be `{"Id": "Val"}` in JSON, but RON produces a one-element tuple instead. The only way to identify that case that I could think of is checking for commas, which unfortunately makes deserializing somewhat quadratic. It could be linear if this was done as a preprocessing pass.

Is there a way to first read a tuple but then change it if it only has one element?

## Failing test case

One of the tests (https://github.com/ron-rs/ron/blob/25aa1eba9bbe39ef777a3d7a1fe2f3c0fb58696d/src/de/value.rs#L256) now fails because it doesn't expect a map. I think it should expect a map, but I don't understand that test's purpose very well, so maybe someone else can comment on this?

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -104,19 +104,84 @@ impl<'de> Deserializer<'de> {
         // Create a working copy
         let mut bytes = self.bytes;
 
-        if bytes.consume("(") {
-            bytes.skip_ws()?;
+        // The caller checks for a brace before calling this
+        if !bytes.consume("(") {
+            unreachable!();
+        }
 
-            if bytes.check_tuple_struct()? {
-                // first argument is technically incorrect, but ignored anyway
+        bytes.skip_ws()?;
+
+        let id = bytes.identifier();
+        bytes.skip_ws()?;
+
+        if id.is_ok() && bytes.peek() == Some(b':') {
+            // first two arguments are technically incorrect, but ignored anyway
+            self.deserialize_struct("", &[], visitor)
+        } else {
+            let mut braces = 1;
+            let mut comma = false;
+            while braces > 0 {
+                let c = bytes.eat_byte()?;
+                if c == b'(' || c == b'[' || c == b'{' {
+                    braces += 1;
+                } else if c == b')' || c == b']' || c == b'}' {
+                    braces -= 1;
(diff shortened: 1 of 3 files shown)


src/de/mod.rs
```diff
@@ -104,21 +104,86 @@ impl<'de> Deserializer<'de> {
         // Create a working copy
         let mut bytes = self.bytes;
 
-        if bytes.consume("(") {
-            bytes.skip_ws()?;
+        // The caller checks for a brace before calling this
+        if !bytes.consume("(") {
```
kvark, line 108: If it's unreachable, why are we checking for it?
If we aren't sure, we should put an assert in. If we are sure, just don't check.

joonazan: Fixed it. There is probably a lot more that can be made more beautiful.

My changes are basically just a proof for myself that fixing this is possible but it ended up working so I'm already using this code myself.

halvnykterist:
I came upon this issue and tried using the PR branch in my code. It seems to work as intended, but there are two issues.
- Line/Column information seems to be missing on the errrors.
- In the case of errors in self-referential enums, a stack overflow (!!) occurs. I've got a minimal example below:
```rust
use serde::Deserialize;
use ron;

#[derive(Deserialize)]
#[serde(untagged)]
enum Number {
    Number(i32),
    SelfReferencing(Box<Number>),
}

fn main() {
    let _: Number = ron::de::from_str("This causes a stack overflow").unwrap();
}
```

kvark:
@torkleyy could you have a look, please?

joonazan:
@halvnykterist This PR should not be merged. Instead, a new version should be written that implements this in a cleaner way.

Maybe there is some way to use serde that I couldn't find. My first post describes the problems I faced.

If not, the best solution would be to feed the input characters into a state machine as they are read in. The current implementation checks the same characters multiple times.

Issue #249, opened 2020-06-23
Suggestion: use a more idiomatic builder pattern in PrettyConfig


allsey87:
Just wanted to point out that I think the use of `with_xxx` methods that take ownership of an instance of `ron::ser::PrettyConfig` and return a modified version of it is not the most idiomatic way of doing things in Rust. As I understand, generally the `with_xxx` methods should create new instances and not modify existing ones...

kvark:
Thank you for filing this!
> with_xxx methods should create new instances and not modify existing ones

Why? The existing ones are consumed anyway. What's the harm in modifying them?

allsey87:
There is no real harm here, it is just a bit unusual when contrasted against other libraries in the Rust ecosystem. For example `Vec::with_capacity` does not modify an existing vector, it creates a new one. The same goes for `HashMap::with_hasher` and `HashMap::with_capacity_and_hasher`. The obvious flaw here is that this doesn't scale well, however, the solution is normally then to use the builder pattern. Taking an example from the [clap](https://docs.rs/clap/2.33.1/clap/) create, they use `Arg::with_name(&str)` and then further configure the returned instance using `.short(&str)`, `.long(&str)`, and `.help(&str)`.

None of this is strictly necessary to follow, but I personally really appreciate these idiomatic coding styles that have emerged since it makes Rust code easier to read and write. Food for thought :)

kvark:
> The obvious flaw here is that this doesn't scale well

This isn't obvious just yet. I'd like to see what may go wrong with the approach we have, in order to motivate changing it.

Currently, since every builder method affects exactly 1 field, and that field is being mutated by the implementation, I don't see what can go wrong. Seems very straightforward and non-prone to errors.

kvark:
Idiomatic code in Rust doesn't appear out of nowhere, it emerges from many use patters, and it's goal is to help code producers and consumers. So in most, if not all, cases where there is an "idiomatic" way, it's based on very practical concerns, they just don't necessarily show themselves as obvious.

In this case, I'm trying to see what practical benefits would this change get us, and so far I wasn't able to see any.

Also, `Vec::with_capacity` and `HashMap::with_capacity_and_hasher` are not relevant, they are just constructors, they aren't using the builder pattern.

The `clap` code has a different kind of builder, it's not chain-able (functions don't return anything). It's hardly more idiomatic.

allsey87:
[...]

PR #259, opened 2020-07-14
move PrettyConfig methods into dedicated builder struct

SirWindfield:
This PR is a follow-up of #258. It moves the construction of a `PrettyConfig` instance into its own `PrettyConfigBuilder` struct, matching the same idea behind #249.

This is more if a place to discuss this further than a real PR. Tests aren't ported over, I just wanted to convey the idea behind it :) It's based upon #258 to already contain the correct method names.

__Edit:__ From my point-of-view:

- Pro:
  - More idiomic. The current `PrettyConfig` is definitely a builder pattern and this conveys it more explicitly.
  - Uses `&mut self`, allowing to pass the builder by reference to other methods if needed.
- Cons:
  - The earlier approach is fine as well. Works without problems.
  - One more method call (`build()`).
  
Pretty obvious, but this would be a breaking change :)

### DIFF
--- a/examples/encode.rs
+++ b/examples/encode.rs
@@ -40,9 +40,9 @@ fn main() {
     };
 
     let pretty = PrettyConfig::new()
-        .with_depth_limit(2)
-        .with_separate_tuple_members(true)
-        .with_enumerate_arrays(true);
+        .depth_limit(2)
+        .separate_tuple_members(true)
+        .enumerate_arrays(true);
     let s = to_string_pretty(&data, pretty).expect("Serialization failed");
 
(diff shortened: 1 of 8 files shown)


kvark:
Thank you for experimenting with this!
I believe the builder patter with an explicit `build` makes most sense where there are things to do in `build`. For example, an API to create a texture could have a builder object that returns you an actual texture in `build()`.
For our case, there is nothing to do on `build`, and there is no need for a separate builder object either. So all in all, it appears to me that introducing it is not necessary, and would just complicate things a bit.

PR #261, opened 2020-07-14
pin repository to msrv of 1.41.0

SirWindfield:


SirWindfield:
I just saw https://github.com/ron-rs/ron/pull/229, which explicitly tries to keep the version as low as possible. It's up for debate if this PR should be merged or not. I personally don't see any benefit on staying on lower MSRV as using rustup the installation can just be updated anyway. And since I use Ubuntu I always try to stick with their latest version.

I do understand though if people want to stick with lower versions. In this case, the version could be pinned to `1.34.0` instead.

PR #269, opened 2020-08-13
Add to_writer_pretty to the ser module

lachlansneff:
Pretty self-explanatory, just adds a `to_writer_pretty` function.

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -16,6 +16,16 @@ where
     value.serialize(&mut s)
 }
 
+/// Serializes `value` into `writer` in a pretty way.
+pub fn to_writer_pretty<W, T>(writer: W, value: &T, config: PrettyConfig) -> Result<()>
+where
+    W: io::Write,
+    T: Serialize,
+{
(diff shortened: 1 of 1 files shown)


src/ser/mod.rs
```diff
@@ -16,6 +16,16 @@ where
     value.serialize(&mut s)
 }
 
+/// Serializes `value` into `writer` in a pretty way.
+pub fn to_writer_pretty<W, T>(writer: W, value: &T, config: PrettyConfig) -> Result<()>
```
kvark, line 20: I think it should be called differently, since it doesn't return a writer. Maybe `write_pretty`?

lachlansneff: I guess it could be, but I was just copying the name from `to_writer` and then adding `_pretty`, in the same way as `to_string` and `to_string_pretty`. If you think about it in terms of the full path, it's `ron::ser::to_writer_pretty`, which does actually make sense.

kvark: oh, sorry, `to_writer` is also a bad name, uh

Issue #268, opened 2020-08-10
Extension for declaring enums, and deserializing as enum repr


ratmice:
This is kind of a can of worms, and merely a first stab/rough draft at thinking about solutions,
but it did come up in some stuff I am working on.  Not sure whether its worth the added complexity given that most use cases don't actually run into anything like this.

I was thinking trying to work with graph data converted into Ron format from other formats.
Below is a simplified example, where the thing to note is that its using _enum values_ for vertices and edges...

```
Graph(
  vertices: {
      Foo: "Some",
      Bar: "random",
      Baz: "stuff"
  },
  edges: {
      "Foo -> Bar": (Foo, Bar),
      "Bar -> Baz": (Bar, Baz),
      "Baz -> Foo": (Baz, Foo),
  }
)
```

To decode this in a program, I then have to hard code into the program the actual enum.
and the program becomes specific to graphs over that set of vertices.
```
// bunch of derives... 
enum ID {
  Foo,
  Bar,
  Baz
}

struct Graph {
  vertices: HashMap<ID, String>,
  edges: HashMap<String, (ID, ID)>,
}
```

I was thinking it could be nice if you could specify the enum and repr _inside the ron file_,
Have the data checked internally as if it were enums, but deserializing to the repr value.

```
#![enable(enum_repr_declarations_deserialized_as_repr)]
#[repr(u8)]
enum ID {
  Foo,
  Bar,
  Baz
}

Graph(
  vertices: {
      Foo: "Some",
      Bar: "random",
      Baz: "stuff"
  },
  edges: {
      "Foo -> Bar": (Foo, Bar),
      "Bar -> Baz": (Bar, Baz),
      "Baz -> Foo": (Baz, Foo),
  }
)
```

Then modify the code as so:
```
struct Graph {
  vertices: HashMap<u8, String>,
  edges: HashMap<String, (u8, u8)>,
}
```

There are a number of questions about how such a thing works when serializing,
The only thing that comes to mind that seems feasible might be using a specialized deserialization/serialization methods that produce and consume serialization mapping from the repr into and out of the enum.

I've looked a bit and don't *think* i've missed anything which could do this sort of thing without an extension -- by all means let me know if otherwise.

kvark:
Unless you have a very specific case where you know what vertices your graph has, it seems rather unfortunate to hardcode the enum variants in the code. Wouldn't it be simpler if the vertices where just strings?
```ron
Graph(
  vertices: {
      "Foo": "Some",
[...]

PR #270, opened 2020-08-14
enum_repr extension

ratmice:
Here is a first attempt at an implementation of the feature requested in #268
I've tried to reduce the impact on typical integer deserialization by introducing a feature gate for it.

Edit:
I forgot to note that the last commit which adds some stricter checking ends up requiring adding 'static lifetime to T: Num + 'static, in a few places, I can't confess to being familiar with lifetimes on non-reference values, but I couldn't figure out any way to add this check without it.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -75,7 +75,7 @@ You can add this extension by adding the following attribute at the top of your 
 This feature enables RON to declare unitary enums with numerical representations,
 There are a number of limitations to this feature:
 
-```ron
+```rust
 #[repr(u8)]
 enum Utensils {
   Bowl,
@@ -87,13 +87,23 @@ enum Ingredients {
   Sugar,
   EggWhites
 }
+```
 
+Where the rust portion gets passed to `ron::parse::TypeEnv`
+
+```ron
 Recipe("meringue": [([Whisk, Bowl], [EggWhites, Sugar])])
 ```
+
+From there when you pass the TypeEnv parsed above to `ron::de::with_type_env`.
+You can decode the RON above using the compiled rust:
 
 ```rust
 struct Recipe<String, Vec<(Vec<u8>, Vec<u64>)>>;
 ```
+
+There is some support for prepending the rust code block to the RON code block,
+parsing both from the same string.
 
 Known limitations (Not expected to change):
 
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -108,9 +108,8 @@ Known limitations (Not expected to change):
 WIP/TODO:
 
 - [x] Proof of concept
-- [ ] Parsing
+- [x] Parsing
 - [ ] [Custom discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations)
-- [ ] Deserialization
-- [ ] Feature gate?
- It seems wise to feature gate this, as otherwise it may impact numerical deserialization even with the extension disabled.
+- [x] Deserialization
+- [x] Feature gate (enum-repr-extension)
 - [ ] Serialization?
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -65,3 +65,52 @@ Enabling the feature would automatically infer `Some(x)` if `x` is given. In thi
     value: 5,
 )
 ```
+
+# enum_repr
+
+You can add this extension by adding the following attribute at the top of your RON document:
+
+`#![enable(enum_repr)]`
+
+This feature enables RON to declare unitary enums with numerical representations,
+There are a number of limitations to this feature:
+
+```ron
+#[repr(u8)]
+enum Utensils {
+  Bowl,
+  Whisk
+}
+
+#[repr(u64)]
+enum Ingredients {
+  Sugar,
+  EggWhites
+}
+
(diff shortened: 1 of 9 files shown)


docs/extensions.md
```diff
@@ -65,3 +65,51 @@ Enabling the feature would automatically infer `Some(x)` if `x` is given. In thi
...
+You can add this extension by adding the following attribute at the top of your RON document:
+
+`#![enable(enum_repr)]`
+
+This feature enables RON to declare unitary enums with numerical representations,
+There are a number of limitations to this feature:
+
+```ron
+#[repr(u8)]
+enum Utensils {
```
kvark, line 80: I'm concerned about this view of the file now. Today, it's just the data. With this change, it becomes a mix of types and data.
I don't think we should mix them. There is no reason why the type declarations have to be specified in the same RON file. I.e. this could be done as a feature of RON library instead of being a RON format extension. For example, we could let the user provide a separate file with ".rs" extension listing those types to be used in RON.
Note: the "rs" extension is because your syntax is 100% a subset of Rust, and it's nothing like RON-described data.

Would you be open to reshaping your feature in this way?

ratmice: Seems like a totally reasonable concern, I believe that doing it as you suggest also probably allows for cleaner the lifetimes associated with the lookup tables i.e. they wouldn't need to be  thread locals any longer.

I want to have a think on it for a day or so, before diving into implementing it though.

ratmice: So, I've done an attempt at splitting up the parsers,
It is by no means a nice patch series to look review at this point.
I'll do a pass cleaning it up once we get the details nailed down.

While working on this I noticed that it is possible to split the parser and have "combined type declarations and ron data",
so they can be parsed either separately, or the types can be *prepended* onto a RON data.

The key to this is that the following is invalid rust since `#![...]` must be at the beginning of a block.

```
enum Foo{}
#![enable(...)]
```

Given the way the parser worked, I figured was actually *more* work to avoid this,
However let me know if you aren't comfortable with that being a thing.

I just wanted to throw it out there as a potential compromise,
since I believe if you wanted to extract Ron code blocks from Markdown,
performing some computed output being able to parse a combined string is going to be much easier with enums inline instead of trying to and trying to correlate
a .rs code block with a .ron one.

Let me know if you still aren't comfortable with that.

kvark: The problem with `enum` declarations in RON is that they are simply too different from the rest. They have different syntax (requiring a different parser), they mess up the extension declaration, like you mentioned:
```
enum Foo{}
#![enable(...)]
```
If we move them to a separate file, it still seems to me that it would solve all the issues:
  1. actual RON files don't even need to know about it! they don't need `#![enable()]` it, there is no need to specify this as a format extension.
  2. the parser code can be completely different. Heck, you can use one of the existing Rust parsers for your definitions, and extra dependencies are fine because you are rolling this as an optional cargo feature anyway
  3. the definition `.rs` is automatically highlighted with Rust syntax in your editor, we don't need to break our heads trying to add support to the extension to our syntax highlighters

ratmice: Alright, well it sounds like it isn't going to particularly fit my needs, I do need the ability to parse these as a single stream, due to embedding in another format which isn't well equipped to correlate the separated and pass the lookup table between the .rs parser and the .ron parser.

As such I think i'll close this and leave the branch up, It will be easier to round out some of the rougher edges with an ad-hoc format by not trying to involve existing formats and syntax anyways.

ratmice: > (sorry for late feedback!)



> (sorry for late feedback!)

No problem, thank you for taking the time to review though!

kvark:
> Alright, well it sounds like it isn't going to particularly fit my needs, I do need the ability to parse these as a single stream, due to embedding in another format which isn't well equipped to correlate the separated and pass the lookup table between the .rs parser and the .ron parser.

Sad news. You want to be able to write markdown stuff as this?
```ron
enum Foo{}
#![enable(...)]
some stuff here
```

And you can't do it as something like:
```rust
enum Foo{}
```
```ron
some stuff here
```

Or maybe even:
```ron
/*! TYPES
enum Foo{}
!*/
some stuff here
```

ratmice:
> > Alright, well it sounds like it isn't going to particularly fit my needs, I do need the ability to parse these as a single stream, due to embedding in another format which isn't well equipped to correlate the separated and pass the lookup table between the .rs parser and the .ron parser.
> 
> Sad news. You want to be able to write markdown stuff as this?
> 
> ```
> enum Foo{}
> #![enable(...)]
> some stuff here
> ```
> 
> And you can't do it as something like:
> 
> ```rust
> enum Foo{}
> ```
> 
> ```
> some stuff here
> ```
> 
> Or maybe even:
> 
> ```
> /*! TYPES
> enum Foo{}
> !*/
> some stuff here
> ```

The thing i'm going for is:

```foo.typed_ron
enum Foo {
  A, B, C
}
#[enable(enum_repr_extension)]
GraphDrawing(Graph(...))
```

Which turns it into a [foo.typed_ron](drawing.png) inline image, or at least the ability to support that case.
Embedding it into a comment *would* work, but then we've got e.g. doc-comments in .rs files extracting markdown, including with ron with rust types embedded in ron comments!

It's just not ergonomic anymore that level of nesting.

PR #287, opened 2020-11-23
Fix overflow check for Number as From<u64>

### LINKED ISSUE #285: Large u64 values overflow as negative i64
Large u64 values that do not fit in a i64 should be stored as f64, by this code:

https://github.com/ron-rs/ron/blob/5e0f3c5e00732c43b457ff10ffb1605321a7b9d0/src/value.rs#L255-L261

However, `i as i64 as u64 == i` is always true. A value that doesn't fit in a a`i64` just overflows in both casts, and results in the same value.

This results in large u64 values being silently converted to large negative i64 values.

torkleyy:
Now big `u64` integers will correctly be mapped to a `f64`.

Fixes #285

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -248,12 +248,12 @@ impl From<i32> for Number {
     }
 }
 
-// The following number conversion checks if the integer fits losslessly into an i64, before
+// The following number conversion checks if the integer fits into an i64, before
 // constructing a Number::Integer variant. If not, the conversion defaults to float.
 
 impl From<u64> for Number {
     fn from(i: u64) -> Number {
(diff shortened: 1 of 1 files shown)


torkleyy:
So while this is clearly an improvement and fixes the obvious bug, note that there's still an issue when trying to round-trip large u64 integers using `Value`.

src/value.rs
```diff
@@ -494,6 +494,8 @@ mod tests {
     {
         use crate::de::from_str;
 
+        dbg!(s);
```
kvark, line 497: leftover stuff

PR #288, opened 2020-11-23
Fix large integer overflow

### LINKED ISSUE #285: Large u64 values overflow as negative i64

JakubValtar:
Hey!

This PR fixes #285.

Not sure if the test is in the right file, let me know if I should move it : )

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -253,7 +253,7 @@ impl From<i32> for Number {
 
 impl From<u64> for Number {
     fn from(i: u64) -> Number {
-        if i as i64 as u64 == i {
+        if i <= std::i64::MAX as u64 {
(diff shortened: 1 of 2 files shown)


torkleyy (approved):

torkleyy:
Thank you!

bors r+

Issue #298, opened 2021-03-11
Add more documentation


hube12:
Hi, 
Since I tinkered a lot with how to do my specific stuff, I have to write this issue, I will try to describe what I wished to achieve and what I managed to do, this would be a great starting point for people that like nice dual way serialization/deserialization. At the **end** I will also mentions current improvement to the pretty config (please **READ** them), I would love a page dedicated to those specific in the wiki or discussion page (you can take this whole article if you want, I give you all the right on it).

My use case is pretty simple, I have a config file I made manually and I need to load it in a specific structure, then add fields and populate it as it goes then write it back. But since it is half/half manual/automatic work, I need to work with a standard format on both side. However my structure is quite big and doesn't have all its fields everytime, but having such in my config would be tiresome, so I had to jump into serde! Let's take a real use case:

```rust
use serde::{Serialize, Serializer, Deserialize, Deserializer};
use std::collections::HashMap;

#[derive(Debug, Deserialize, Serialize)]
pub struct Game {
    macros: HashMap<String, Macros>, // those macros are applied to the actions
    actions: HashMap<String, String>, // in this version the actions are not described for conciseness
}

#[derive(Debug, Deserialize, Serialize)]
pub struct Macros {
    superseded: Option<String>, // this is the macros that is extended
    subject: SubjectMapping, // we want to force a subject even if the subject has no element
    action: Option<ActionMapping>, // we don't really care if there is an action
    how: Vec<String>, // this is either empty or filled
}

#[derive(Debug, Deserialize, Serialize)]
pub struct SubjectMapping {
    primary: Option<String>,
    secondary: Option<String>,
}

#[derive(Debug, Deserialize, Serialize)]
pub struct ActionMapping {
    primary: Option<String>,
}
```
game.ron :
```ron
(
    macros:{
       "defaults" : (
           action:()
       ),
       "hero" : (
           superseded: "defaults",
           subject: (
             primary: "hero",
             secondary: "paladin",
           ),
           action:(
             primary: "attack"
           ),
           how: ["sword","magic"],
       ),
    },
    actions:{}
)
```

As we can see our structure is semi filled, so if we try to deserialize it we will face some issues (I will use here include_str but as this is a dynamic asset you want to use a bufreader and from_reader.

```rust
#[test]
fn test_deserialize() {
    const GAME: &str = include_str!("../constants/game.ron");
    use ron::from_str;
    let r: Result<Game, _> = from_str(GAME);
    dbg!(&r);
[...]

PR #299, opened 2021-03-12
feat(add compact arrays): we can now have nice one line syntax for arrays

hube12:
This refer to my recent issue #298 where I mention I would love a compact array definition,
This PR allow to do so by using the `with_compact_arrays(true)`, the default is false ofc.
```rust
assert_eq!(
        to_string_pretty(&&[(), (), ()][..], PrettyConfig::new().with_new_line("\n".to_string()).with_compact_arrays(true)).unwrap(),
        "[(),(),()]"
    );
```

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -733,7 +733,21 @@ impl<'a, W: io::Write> ser::SerializeSeq for Compound<'a, W> {
             }
             Compound::Map { ser, .. } => ser,
         };
-        ser.end_indent()?;
+
+        if let Some((ref config, ref mut pretty)) = ser.pretty {
+            if pretty.indent <= config.depth_limit {
+                let is_empty = ser.is_empty.unwrap_or(false);
+
+                if !is_empty && !config.compact_arrays {
+                    for _ in 1..pretty.indent {
+                        ser.output.write_all(config.indentor.as_bytes())?;
+                    }
+                }
+            }
+            pretty.indent -= 1;
+
+            ser.is_empty = None;
+        }
 
         if let Some((_, ref mut pretty)) = ser.pretty {
             pretty.sequence_index.pop();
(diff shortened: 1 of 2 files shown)


src/ser/mod.rs
```diff
@@ -520,7 +544,16 @@ impl<'a, W: io::Write> ser::Serializer for &'a mut Serializer<W> {
             self.is_empty = Some(len == 0);
         }
 
-        self.start_indent()?;
+        if let Some((ref config, ref mut pretty)) = self.pretty {
```
kvark, line 547: can this be done more flexible? For example,
```rust
fn start_indent_if(&mut self, condition: impl Fn(&Config) -> bool) {
...
}
fn start_indent(&mut self) {
  self.start_indent_if(|_| true)
}
```
And the same for `fn indent()` and `fn end_indent()`. This would allow us to re-use the code better.

hube12: Yeah that was exactly what I though but I didn't wanted to make signature change without some people agreeing first.

hube12:
I think it has good as it can gets for this specific use case, there is still the issue of structs inside array so maybe compact_arrays could be extended to compact_array_if_level_below(level) so only leaf structure are compact but nested ones aren't

torkleyy:
> 
> 
> I think it has good as it can gets for this specific use case, there is still the issue of structs inside array so maybe compact_arrays could be extended to compact_array_if_level_below(level) so only leaf structure are compact but nested ones aren't

Either that, or define a char limit for "compact arrays" (could do the same with structs / maps, too, I guess)

kvark (approved):

kvark:
Here are 2 things that we'll need to follow-up with:
  1. same PR as this for the `master` branch
  2. add a commit to this PR with a patch version bump and a short CHANGELOG entry

kvark:
comment

MomoLangenstein:
The `compact_arrays` option has now been included in #349

PR #304, opened 2021-05-13
make u128, i128 behind feature integer128

clouds56:
for some target that not support u128 like emscripten

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,7 @@ serde = { version = "1.0.60", features = ["serde_derive"] }
 [dev-dependencies]
 serde_bytes = "0.11"
 serde_json = "1"
+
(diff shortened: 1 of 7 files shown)


kvark:
Sorry for getting late here. A feature has costs in that it needs to be tested, and as features grow the testing becomes more difficult. So generally we'd want to avoid features where possible.
Here, if some platforms don't support 128-bit integers, could we instead detect them via `cfg` flags?

clouds56:
Yes, we could do this just like what it does in serde.
But I don't think it would make test easier since we have to install another compiler for test.

kvark:
Is this particular thing difficult to test on CI? I.e. does github CI have any platform targets that don't have 128-bit scalars?

PR #308, opened 2021-06-30
Add fuzzer from &str to ron::Value

5225225:
The fuzzer requires nightly to build, so to run it you need `cargo +nightly fuzz run from_str`.


kvark (approved):
Amazing stuff, thank you!
bors r+

kvark:
Sorry about late review! Looks like MSRV isn't happy. Do we have a way to make it happy, or do we need to bump MSRV here?

5225225:
On 6 July 2021 14:51:24 UTC, Dzmitry Malyshau ***@***.***> wrote:
>Sorry about late review! Looks like MSRV isn't happy. Do we have a way to make it happy, or do we need to bump MSRV here?
>

Well, running the fuzzer is nightly only, but that shouldn't affect anyone who wants to use ron as a dependency.

Does cargo test try to build the fuzzer target? It *shouldn't, and I didn't touch any files outside of the fuzzer dir.

Currently away from my computer at the moment but when I get home I'll have a look at it. My guess is it'll be some form of flag passed to the cargo test to make it ignore that fuzz directory.

kvark:
Ok, I see. We just need to modify CI scripts to not run the fuzzer on MSRV.

5225225:
`master` doesn't respect MSRV because `indexmap` needs nightly. Is that feature exempt from the MSRV? You'll probably need 2 tests, one that runs with nightly and `--all-features`, one that runs with the full matrix, but without `--all-features.

Maybe fix that on a different PR and then retry the bors here?

kvark:
Why does indexmap require nightly? https://github.com/bluss/indexmap says 1.49 is required.

5225225:
Ahh, you're right, never mind. I tested with `cargo +stable test --all-features` and that works fine. I was getting a nightly feature error by default, forgot that it was stabilized since then.

5225225:
Hey, sorry for forgetting about this PR. Just rebased ontop of `master` (though there were no conflicts).

I ran the commands that exist in the `ci.yaml` to downgrade the packages, and it seems to work fine.

torkleyy (approved):
Thank you! :shipit:

PR #310, opened 2021-08-07
Add the Extension `implicit_outmost_struct`

AnarchY1n:


src/de/mod.rs
```diff
@@ -483,6 +483,9 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         self.bytes.skip_ws()?;
 
         if self.bytes.consume("(") {
+            if self.bytes.exts.contains(Extensions::IMPLICIT_OUTMOST_STRUCT) {
```
kvark, line 486: We are in a general struct deserialization routine here. If we error whenever `IMPLICIT_OUTMOST_STRUCT` is enabled, we wouldn't be able to deserialize any non-root struct. Correct?

kvark (changes requested):
Thank you for the PR, and sorry about the late review!

PR #311, opened 2021-08-08
Fix some clippy warnings

remilauzier:


kvark (approved):
That's neat!

kvark:
uh, the CI seems to complain about something else

PR #319, opened 2021-10-13
Adds unwrap_variant_newtypes extension

### LINKED ISSUE #250: Avoiding duplicate parens around enum variant members (related to newtype-wrapping)
E.g. if I have this:
```rust
#[derive(Deserialize)]
pub enum Foo {
	A(A),
	B(f32),
}

#[derive(Deserialize)]
pub struct A {
	pub b: u8,
}
```
in RON I have to write `A(( b: 42 ))`. 
It would be preferable to be able to just write `A( b: 42 )` (similar to the already implemented [newtype-wrapping](https://github.com/ron-rs/ron/pull/72)), but currently this gives an error.

juntyr:
This PR fixes #250 by adding a new `unwrap_variant_newtypes` extension. When enabled, newtype variants will skip the outer parentheses of the wrapped type:
```rust
#[derive(Deserialize)]
pub enum Foo {
	A(A),
	B,
}

#[derive(Deserialize)]
pub struct A {
	pub b: u8,
}
```
In this example, `Foo` can now be parsed from `A(b: 4)`.

### Notes

* By adding a new feature, no breaking change occurs.
* By enabling the `unwrap_variant_newtypes`, neither `A(A(b: 4))` nor `A((b: 4))` will parse, as unwrapping newtype variants is always performed.
* This extension can mimic some of the functionality of `unwrap_newtypes` if the newtype variant wraps around a newtype struct. Both extensions are compatible, but in this case `unwrap_newtypes` is not needed as the newtype variant can also perform the unwrapping.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -65,3 +65,43 @@ Enabling the feature would automatically infer `Some(x)` if `x` is given. In thi
     value: 5,
 )
 ```
+
+# unwrap_variant_newtypes
+
+You can add this extension by adding the following attribute at the top of your RON document:
+
+`#![enable(unwrap_variant_newtypes)]`
+
+This feature enables RON to automatically unwrap newtype enum variants.
+
+```rust
+#[derive(Deserialize)]
+struct Inner {
+    pub a: u8,
+	pub b: bool,
+}
+#[derive(Deserialize)]
+pub enum Enum {
+	A(A),
+	B,
+}
+```
+
+Without `unwrap_variant_newtypes`, your RON document would look like this:
+
+``` ron
+(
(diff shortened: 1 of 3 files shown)


torkleyy:
Thanks for the explanation! @MomoLangenstein 
That certainly makes sense. What do you think about not allowing to write out the newtype with the extension enabled?

MomoLangenstein:
> 
> 
> Yes, exactly @MomoLangenstein, that's what I was thinking. If you can do that, I'd be willing to merge this - except @kvark has any concerns?

Brilliant, I'll change up the PR then

Issue #264, opened 2020-08-04
Extracting struct name from Value?


jgarvin:
Looking at the docs for Value:

```
pub enum Value {
    Bool(bool),
    Char(char),
    Map(Map),
    Number(Number),
    Option(Option<Box<Value>>),
    String(String),
    Seq(Vec<Value>),
    Unit,
}
```

I don't see how to get the name of a struct out, e.g. if I were to call `from_str` on the contents of `example.ron`, how would I get the string "Nested" out of it? My use case is the following: I want to have a field that could be one of many different structs. I'd like to parse into a ron Value, check the parsed struct name, then based on that name choose which actual `T` to pass to `into_rust`. I couldn't find this in the examples/docs.

jgarvin:
So I spent some time playing around with adding:

```
`#[derive(Clone, Debug, Default, Deserialize, Serialize, Eq)]
pub struct Structure {
    pub name: String,
    pub fields: Map,
}

impl Hash for Structure {
    fn hash<H: Hasher>(&self, state: &mut H) {
        // TODO: is this any different than #derive'ing Hash?
        self.name.hash(state);
        self.fields.hash(state);
    }
}

impl Ord for Structure {
    fn cmp(&self, other: &Structure) -> Ordering {
        (&self.name, &self.fields).cmp(&(&other.name, &other.fields))
    }
}

/// Note: equality is only given name matches and if both values and order of values match
impl PartialEq for Structure {
    fn eq(&self, other: &Structure) -> bool {
        self.name == other.name && self.fields == other.fields
    }
}

impl PartialOrd for Structure {
    fn partial_cmp(&self, other: &Structure) -> Option<Ordering> {
        (&self.name, &self.fields).partial_cmp(&(&other.name, &other.fields))
    }
}

#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
pub enum Value {
    Bool(bool),
    Char(char),
    Structure(Structure),
    Map(Map),
    Number(Number),
    Option(Option<Box<Value>>),
    String(String),
    Seq(Vec<Value>),
    Unit,
}
```

But the problem I'm running into is that serde's data model seems to force losing struct names when combined with the lack of a parser separate from serde. What I see:

* parse.rs doesn't give a full parser, it appears to be utility functions for implementing one
* Deserializer actually contains the parser, and uses the parse.rs functions.
* Deserializer needs you to map the input onto the serde data model.
[...]

Issue #189, opened 2019-10-07
to_string(from_str::<Value>("ron")) should be lossless


CAD97:
The specific case I'm running into that fails to round trip is serializes via [`Serializer::serialize_newtype_variant`](https://docs.serde.rs/serde/trait.Serializer.html#tymethod.serialize_newtype_variant). I believe that this is because `Value` doesn't store the name of the struct stored, so deserialized named things become unnamed sequences.

cc #122

pheki:
I'm not sure this is possible with the current serde::Serializer trait, as `serialize_newtype_variant` takes a `&'static str` as name, which we can't get when deserializing (unless we leak a String).

I also made a simple test for the same problem but with Struct instead:

```
use ron::Value;

#[test]
fn struct_round_trip() {
    let ron_str = "GameConfig(field: 2)";
    let value: Value = ron::from_str(ron_str).unwrap();
    let reserialized = ron::to_string(&value).unwrap();
    assert_eq!(ron_str, reserialized);
}
```

It fails with `"GameConfig(field: 2)" != "({\"field\":2})"`

It also seems to me that the deserialization step would be impossible to implement following the abstraction proposed by serde in which Value would be independent of the serializer as Visitor does not have a `visit_struct` method, so there's no way to communicate to the visitor that its visiting a struct, only a map (see how deserialize_struct calls visit_map [here](https://github.com/ron-rs/ron/blob/3984332e191b9eb1f12b2a11fc7dd92cf56387e9/src/de/mod.rs#L486)).

Maybe there could be a hack by coupling `ron::Value` to `ron::Deserializer` and `ron::Serializer`.

PR #328, opened 2021-10-26
Add Value::Struct variant and preserve struct names and formatting

cswinter:
This change adds a ~dreadful hack~creative workaround that allows us to distinguish between structs and maps during parsing and preserve struct names. It does so by abusing the `size_hint` method of the `MapAccessor` as a side channel that exposes information about the struct name and whether we are visiting a struct or a map. I am somewhat confident that, within the context of the current implementation, this is sound and won't unexpectedly blow up in our face. The `CommaSeparated` struct with the unusual `size_hint` method is passed directly to the `ValueVisitor` and drained before we return control flow to any serde code, so anything else that might call the `size_hint` method will just receive `None` values. I haven't confirmed this, but with some luck, `size_hint` will also get inlined after monomorphization and optimized into something efficient.

This makes progress towards https://github.com/ron-rs/ron/issues/122, fixes https://github.com/ron-rs/ron/issues/189, and fixes https://github.com/ron-rs/ron/issues/264.

This change has been arrived at by sheer power of will rather than any kind of understanding of serde so there may be a better solution. There are still some smaller issues with the PR, before fixing those I first wanted to check whether this has any hope of getting merged.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -1,7 +1,6 @@
 /// Deserialization module.
 pub use crate::error::{Error, ErrorCode, Result};
 pub use crate::parse::Position;
-use crate::{Map, Value};
 
 use serde::de::{self, DeserializeSeed, Deserializer as SerdeError, Visitor};
 use std::cell::RefCell;
--- a/src/error.rs
+++ b/src/error.rs
@@ -106,7 +106,6 @@ impl fmt::Display for ErrorCode {
             ErrorCode::UnderscoreAtBeginning => f.write_str("Found underscore at the beginning"),
             ErrorCode::UnexpectedByte(_) => f.write_str("Unexpected byte"),
             ErrorCode::TrailingCharacters => f.write_str("Non-whitespace trailing characters"),
-            _ => f.write_str("Unknown ErrorCode"),
         }
     }
 }
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -1,6 +1,7 @@
 use serde::{ser, Deserialize, Serialize};
 use std::io;
 
+use crate::Value;
 use crate::{
     error::{Error, Result},
     extensions::Extensions,
@@ -56,6 +57,22 @@ where
     Ok(String::from_utf8(s.output).expect("Ron should be utf-8"))
 }
 
+impl Value {
+    pub fn to_string(&self) -> Result<String> {
+        let buf = Vec::new();
+        let mut s = Serializer::new(buf, None)?;
+        self.enhanced_serialize(&mut s)?;
+        Ok(String::from_utf8(s.output).expect("Ron should be utf-8"))
+    }
+
+    pub fn to_string_pretty(&self, config: PrettyConfig) -> Result<String> {
+        let buf = Vec::new();
+        let mut s = Serializer::new(buf, Some(config))?;
+        self.enhanced_serialize(&mut s)?;
+        Ok(String::from_utf8(s.output).expect("Ron should be utf-8"))
+    }
+}
+
 /// Pretty serializer state
 struct Pretty {
     indent: usize,
@@ -86,7 +103,7 @@ pub struct PrettyConfig {
     /// Indentation string
     #[serde(default = "default_indentor")]
     pub indentor: String,
-    // Whether to emit struct names
+    /// Whether to emit struct names
     #[serde(default = "default_struct_names")]
     pub struct_names: bool,
     /// Separate tuple members with indentation
@@ -368,6 +385,21 @@ impl<W: io::Write> Serializer<W> {
             .map(|(pc, _)| pc.struct_names)
             .unwrap_or(false)
     }
+
+    fn serialize_struct_dyn<'b>(&mut self, name: &'b str, len: usize) -> Result<Compound<W>> {
+        if self.struct_names() {
+            self.write_identifier(name)?;
+        }
+        self.output.write_all(b"(")?;
+
+        self.is_empty = Some(len == 0);
+        self.start_indent()?;
+
+        Ok(Compound {
+            ser: self,
+            state: State::First,
+        })
+    }
 }
 
 impl<'a, W: io::Write> ser::Serializer for &'a mut Serializer<W> {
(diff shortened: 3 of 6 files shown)


kvark:
Sorry, I've been punting on reviewing this.
@torkleyy is this something you'd be comfortable reviewing?

cswinter:
So I did get this to work and am using it in https://github.com/cswinter/pyron to (de)serialize Python dataclasses as ron, but there are still some issues with this that I'm not sure yet if they can be fixed. Both the `ValueVisitor` used in the `Deserialize` implementation of `Value`, and the `Deserializer` fail to conform to the expected `serde` interface and are public which might cause some issues if you use them in combination with a different `Deserialize`/`Deserializer`.
Particularly, if you try to deserialize something else (e.g. yaml) into a `ron::Value`, the `ValueVisitor` won't behave correctly and do weird things if it receives size hints. We could possibly fix this by further changing the ron `Deserializer` to emit a sentinel value of `usize::MAX` as the first `size_hint` (which would never be emitted by a valid `Deserializer`) and allows the `ValueVisitor` to work normally if it doesn't see this sentinel. However, this would now break the `Deserializer`, and actually, my current implementation might already be broken for deserializing into structs. There could still be some way around this but I'm not sure yet if that's the way to go. It might be better to just write a new parser from scratch that is not tied to serde, which I'm considering anyway since it enables some other features that are not currently possible.

The serialization implementation that allows preserving struct names/tuples is in better shape. It requires some amount of code duplication, and unfortunately needs an explicit call to a new `Value::to_string/to_pretty_string` method so we can call our custom serialize method that doesn't quite work like the `serde` one, but I'm not away of any glaring soundness issues. It's a lot less useful with the corresponding deserialization code of course.

cswinter:
For now, my main use case for this is https://github.com/cswinter/pyron, which doesn't use ron in a way that would run into the remaining issues. I would like to fix this properly at some point but still need to decide on the right approach and exactly what features I want so I'm planning to just let this stew for a while. The main approaches I'm considering:
- come up with one more clever idea to make this hack work
- implement custom deserializer (and maybe also pretty printer) which isn't tied to serde
- get `serde` to add some new methods for us that allow an implementation without hacks (I think this could be done in a way that doesn't break existing methods by adding alternative variants with dynamic lifetime etc. which have a default implementation that returns a specific error. Existing serializer would just never call these, and our serializer could call these to use the extra functionality when available, or revert to the current behavior for `Deserializer`/`Deserialize` implementations that don't support this functionality.)

If anyone has any other good ideas or opinions on how this should be approached I'd be glad to hear about them.

torkleyy:
> 
> 
> Sorry, I've been punting on reviewing this. @torkleyy is this something you'd be comfortable reviewing?

I might be able to next week, feel free to ping me if I forget.

kvark:
@torkleyy ping

juntyr:
I've been thinking about a more AST-like Value type again and this is a sketch of what a `NewValue` might look like:

```rust
pub enum NewValue<S: std::borrow::Borrow<str>> {
    Unit,
    Bool(bool),
    Number(Number),
    Char(char),
    String(String),
    Bytes(Vec<u8>),
    List(Vec<NewValue<S>>),
    Map(NewMap<S>),
    NamedUnit {
        name: S,
    },
    Newtype(Box<NewValue<S>>),
    NamedNewtype {
        name: S,
        inner: Box<NewValue<S>>,
    },
    Tuple(Vec<NewValue<S>>),
    NamedTuple {
        name: S,
        elems: Vec<NewValue<S>>,
    },
    Struct {
        fields: FieldMap<S>,
    },
    NamedStruct {
        name: S,
        fields: FieldMap<S>,
    }
}

#[cfg(not(feature = "indexmap"))]
type NewMap<S> = std::collections::BTreeMap<NewValue<S>, NewValue<S>>;
#[cfg(feature = "indexmap")]
type NewMap<S> = indexmap::IndexMap<NewValue<S>, NewValue<S>>;

#[cfg(not(feature = "indexmap"))]
type FieldMap<S> = std::collections::BTreeMap<S, NewValue<S>>;
#[cfg(feature = "indexmap")]
type FieldMap<S> = indexmap::IndexMap<S, NewValue<S>>;
```

I think the biggest questions are:

* how to represent `Option`, since the unwrap newtype variants feature includes `Some` and thus all of the following are allowed: `None`, `Some(val)`, `Some()` = `Some(())`, `Some(a, b)` = `Some((a, b))`, and `Some(a: a, b: b)` = `Some((a: a, b: b))`. On a syntactic level, `None` is just a named ident and `Some` is just a named newtype / tuple / struct with special handling on serde's side
* how to handle extensions - maybe we could support attributes more generally around every value subtree?
* how to handle newtypes, since ron sometimes erases them

rlidwka:
> I've been thinking about a more AST-like Value type again

Is there any progress made on this?

I'd like to have lossless Value, which I can have multiple attempts at trying to deserialize from for untagged enums. Trying to make something like this work:

```rs
#[derive(Debug, serde::Serialize, serde::Deserialize)]
enum A {
    A1(String),
    A2(String),
}

#[derive(Debug, serde::Serialize, serde::Deserialize)]
enum B {
    B1(String),
    B2(String),
}

#[derive(Debug, serde::Serialize, serde::Deserialize)]
#[serde(untagged)]
enum MyEnum {
    A(A),
    B(B),
}

let value = MyEnum::A(A::A1("Hello, world!".to_string()));

// valid RON serializations are:
// A1("")
// A2("")
// B1("")
// B2("")

// works fine with serde_json
let json = serde_json::to_string_pretty(&value).unwrap();
dbg!(serde_json::from_str::<MyEnum>(&json).unwrap());

// but doesn't work with ron
let ron = ron::ser::to_string_pretty(&value, ron::ser::PrettyConfig::default()).unwrap();
dbg!(ron::de::from_str::<MyEnum>(&ron).unwrap());
```

juntyr:
> Is there any progress made on this?

Not yet, unfortunately. Once I am back from parental leave, I will first tackle *finally* publishing v0.9, which includes a list of all the weird serde attribute cases that we currently cannot support because serde treats every format like JSON. If serde were to ever change that (and e.g. impl Trait in type aliases in traits makes that technically possible), an improved version of Ron’s Value type (as outlined above) would be used to fix most of these cases.

While it will take me a few more months to get back to this PR, feel free to implement something similar to my idea sketched out above (essentially an AST-like Value type that allows round trips of RON [without serde attributes] but is more an AST than a Value enum since e.g. named structs and enum variants are ambiguous in RON when we are not given type information by the user).

Issue #200, opened 2019-12-19
How to specify whether to emit optional struct name during serialization?


Boscop:
How to specify whether to emit optional struct name during serialization?

I see no option for that here
https://docs.rs/ron/0.5.1/ron/ser/struct.PrettyConfig.html

but I think there should be one..

kvark:
We've been discussing this in https://github.com/ron-rs/ron/issues/105#issuecomment-567256298 and the recent conclusion is that the bool must be moved into `PrettyConfig`.
Also, since we are at it, would be great to use the `#[non_exhaustive]` on the config, which just became available in stable rust.

If you want to take it, please do!

PR #329, opened 2021-10-28
Move option to emit struct names to `PrettyConfig`

cswinter:
Fixes https://github.com/ron-rs/ron/issues/200.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+- Add `struct_names` option to `PrettyConfig`
+
 ## [0.7.0] - 2021-10-22
 
 - Add `unwrap_variant_newtypes` extension ([#319](https://github.com/ron-rs/ron/pull/319))
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -17,7 +17,7 @@ where
     W: io::Write,
     T: ?Sized + Serialize,
 {
-    let mut s = Serializer::new(writer, None, false)?;
+    let mut s = Serializer::new(writer, None)?;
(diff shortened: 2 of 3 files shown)


kvark (approved):
Very nice! Also, I believe this is backwards-compatible.
Edit: Actually, no, it's breaking `pub fn new(..)` API.

malte-v:
I actually think that this is a bad idea and should be reverted. Emitting struct names is not only a matter of prettiness. Let's say I have two structs, `Foo` and `Bar`:
```
struct Foo { inner: u8 }
struct Bar { inner: u8 }
```
When serializing a `Foo` and a `Bar` with the same `inner` value, their serialized forms will look the same if struct names are skipped. As a consequence, their cryptographic hashes and signatures will also be the same, so an attacker could take a signed `Foo` and present it as a signed `Bar`, even though the two structs may represent totally different information.

kvark:
Sorry,I don't understand the argument about cryptographic names and hashes.
You want different types to look different in the serialized form? That's already not the case. `1u8` and `1u32` would look exactly the same.

malte-v:
The fact that `1u8` and `1u32` look the same in serialized form is not a problem because they are semantically the same: They're just integers. Compound types, on the other hand, are often structurally the same but semantically different. That's where emitting struct names comes in handy.

Let me give a more concrete example: you have a kind of decentralized business review platform where each customer has its own keypair and signs reviews:
```
struct GoodReview(id: BusinessId);
struct BadReview(id: BusinessId);
```
If you just serialize, hash and sign the above structs without emitting struct names, any valid `BadReview` will be a valid `GoodReview` and vice versa. If you emitted struct names, this kind of attack would be impossible.

(In the example above, you'd probably just use an enum instead. But if you have more structs to sign and want absolute peace of mind, you'd have to put all your structs into an enum, which is just a pain to maintain and introduces new kinds of possible errors. Emitting struct names is a much simpler and more elegant solution.)

malte-v:
Another more general argument is that since this has been merged, your `PrettyConfig` makes a difference that goes beyond aesthetics: It decides whether you will be able to falsely deserialize into a struct that is structurally the same but not identical to the struct that the value was serialized from.

PR #331, opened 2021-10-29
Fix newtype variant unwrapping inside enum, seq and map

MomoLangenstein:
This is a follow-up to #319 which added the `unwrap_variant_newtypes` extension.

Unfortunately, it seems like I missed a few interactions with parsing enums, sequences and maps nested inside a newtype variant (i.e. not inside a containing non-newtype struct). I have fixed all three and added extra test cases to cover their intended functionality, i.e. `NewtypeVariant( [ Struct(...) ] )`, `NewtypeVariant( { "key": Struct(...) } )`, and `NewtypeVariant( StructVariant(...) )` all parse again.

### Discussion

`Option<T>` is another fun case. Here I think carrying through the unwrapping actually is quite neat, but I want to get your input if this should be the intended behaviour:
* `NewtypeVariant( None )`
* `NewtypeVariant( Some(a: 4, bool: false) )` for `Option<Struct>` and `struct Struct { a: u32, b: bool }`
* `NewtypeVariant( a: 4, bool: false )` is the `implicit_some` extension is switched on as well

I am also a bit unsure about how to handle `deserialize_any`. I think I may need your help with that one.

### TODO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 - Add `struct_names` option to `PrettyConfig`
+- Fix newtype variant unwrapping around enum, seq and map ([#331](https://github.com/ron-rs/ron/pull/331))
 
 ## [0.7.0] - 2021-10-22
 
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -129,6 +129,10 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
+        // Newtype variants can only be unwrapped if we receive information
+        //  about the wrapped type - with `deserialize_any` we don't
+        self.newtype_variant = false;
+
         if self.bytes.consume_ident("true") {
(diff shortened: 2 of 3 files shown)


torkleyy:
Hey, thanks for the follow-up! I'm not sure when I will be able to review this.

> I am also a bit unsure about how to handle deserialize_any. I think I may need your help with that one.

I don't think there's anything special to do for `deserialize_any`; if there's no type information given to us we don't need to implicitly generate anything. That does mean that self-described and statically typed data will not deserialize into the same thing, but that is also the case with other self-describing serde formats (and already the case with RON as well).

MomoLangenstein:
I guess it might be easiest then to just turn off the newtype variant unwrapping if we enter `deserialize_any`? So then the semantics would be that unless type information is given, the unwrapping cannot be performed. Ideally there wouldn’t be a special case, of course, but some types (e.g. wrapped unit and struct) can only be parsed if they are described in the RON, which is exactly what the extension avoids.

MomoLangenstein:
@torkleyy I've disabled newtype variant unwrapping inside `deserialize_any` as discussed and added the fix to the changelog

torkleyy (approved):
Thanks, looks good to me!

Issue #281, opened 2020-11-09
Ability to configure `Deserializer` to enable extensions when parsing


Boscop:
Problem: I want `#![enable(unwrap_newtypes, implicit_some)]` to be the default, and don't want my users to require to write it in every RON file.

Current workaround: Inject that line before parsing. It's suboptimal because I can't use `BufReader` and if there's a parsing error, the reported line number will be wrong from the user's POV.

Proposed Solution: Ability to configure `Deserializer` to enable extensions before parsing.

kvark:
That seems plausible.

torkleyy:
I've been thinking about this issue a lot and my own conclusion is:

## What are RON extensions?

RON extensions, right now, are deserialization options that are configured together with the deserialized data (in the .ron file). They **do not** influence parsing of RON - the syntax always stays the same.

## Considerations

Knowing a Rust type, one might have certain expectations about how to write correct RON for it.

However, that's not all there is to it: What RON is correct and what isn't **is completely controlled by what deserialization code serde comes up with**. If I were to use a different deserialization library I might have to write different RON for it (e.g. it might not accept a RON tuple when a fixed-size array is expected).

Now that means that we have two options to control deserialization:

1. Configuring what the RON deserializer tells serde, in terms of serde's type model (`RON -> serde`, RON extensions)
2. Configuring how serde transforms "serde data" to Rust structures (`serde -> Rust`, serde attributes)

As it turns out, there are many options in `serde` to control deserialization (https://serde.rs/attributes.html). Now, does that mean RON extensions are unnecessary because we can achieve the same thing with serde? Some examples:

### `implicit_some`

Could be replaced by serde attributes by implementing a custom visitor which visits `deserialize_any` and forwards deserialization to the type wrapped in `Option`. I think it would be possible to implement a generic module which can be used like this:

```rust
#[derive(Deserialize)]
struct Foo {
    #[serde(with = "implicit_some")]
    x: Option<i32>
}
```

### other extensions

I was gonna look at each extension, but it seems that the same applies all the other extensions; they could be replaced in a similar fashion.

## Are serde attributes a true replacement?

They are not the same. Some differences:

* Requires access to the code with the derive
* Possibly tedious to add all those attributes
* No longer can the author of the RON file choose for themselves how explicit they want to be in terms of options, newtypes, etc (the developer makes the decision for them)

## Conclusion

[...]

PR #339, opened 2021-11-17
Add `output_extensions` option to `PrettyConfig`

MomoLangenstein:
This PR adds an extra on-by-default configuration option to `PrettyConfig`. When `output_extensions` is set to `false`, the serialized RON will **not** include the `#![enable(EXTENSION)]` attributes.

This would also be related to #281, i.e. for use cases where RON is used e.g. for a configuration file that is both produced and consumed by the same program, extensions could be automatically enabled on deserialization and removed on serialization to remove the overhead.

While programmatically enabling extensions is doable already (just insert the attribute-enabling string before the RON), disabling them is much harder as you would have to parse and match on the attributes' format.

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -100,6 +100,8 @@ pub struct PrettyConfig {
     /// Enable extensions. Only configures 'implicit_some',
     ///  'unwrap_newtypes', and 'unwrap_variant_newtypes' for now.
     pub extensions: Extensions,
+    #[serde(default = "default_output_extensions")]
+    pub output_extensions: bool,
     /// Private field to ensure adding a field is non-breaking.
     #[serde(skip)]
     _future_proof: (),
@@ -191,6 +193,16 @@ impl PrettyConfig {
 
         self
     }
+
+    /// Configures whether the extensions are included in the output as
+    ///  `#![enable(..)]` attributes
+    ///
+    /// Default: `true`
+    pub fn output_extensions(mut self, output_extensions: bool) -> Self {
+        self.output_extensions = output_extensions;
+
+        self
+    }
 }
 
 fn default_depth_limit() -> usize {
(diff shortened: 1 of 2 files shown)


MomoLangenstein:
@torkleyy One last PR (for now) that adds a tiny bit of extra functionality. Due to its relation to #281 I'd be very interested if you think this is the correct design for this feature.

src/ser/mod.rs
```diff
@@ -191,6 +193,16 @@ impl PrettyConfig {
 
         self
     }
+
+    /// Configures whether the extensions are included in the output as
```
torkleyy, line 197: I think we should describe the use-case here and what to look out for (enabling the same extensions when deserializing).

torkleyy (commented):
I was a bit skeptical about this originally, but I think this is the best solution.

However, I think we should provide a safer API to make it less confusing and error prone. I was thinking something like this:

```rust
function ron_options() {
    ron::Options::build().implicit_some().build()
}

let ron = ron_options();

let foo: Foo = ron.from_file("foo.ron")?;
ron.to_file("foo.ron")?; // same options used for deserialization and serialization
ron.to_file_pretty("foo.ron", PrettyConfig::default())?;
```

We should also make sure to provide such a code snippet as documentation and to explain that one has to use the same options for serialization and deserialization. I think doing all this cleanly might involve some breaking API changes, not sure.

MomoLangenstein:
I really like the design you are proposing here. Do you think it would be best to first implement this new API and then after that the functionality to turn off the extension outputting? I might make sense to give a nicer usage example.

torkleyy:
Yes, I think that would be the best approach @MomoLangenstein :+1:

MomoLangenstein:
> 
> 
> Yes, I think that would be the best approach @MomoLangenstein 👍

Ok, I'll write that up then (maybe tomorrow)

MomoLangenstein:
This PR will be superseded by #343 and closed once that is merged

MomoLangenstein:
With #343 this has now become obsolete

Issue #332, opened 2021-10-30
Error building ron with current rust compiler


greg7mdp:
With a newly installed rust, both on windows and ubuntu, I have the following error when trying to build the current version of ron:
![image](https://user-images.githubusercontent.com/1496907/139555575-9416ab96-77a9-4009-9eca-18b1fac713a7.png)

MomoLangenstein:
If you are building from inside `ron`‘s folder, you might be building with its minimum supported Rust version, 1.36.0. What helps me in that case is to follow `ron`‘s CI action and run
```shell
cargo update -p bitflags --precise 1.2.1
cargo update -p indexmap --precise 1.6.2
```

torkleyy:
If you need to build with an old Rust version, yes, you will have to do what @MomoLangenstein posted.

However, if you can @greg7mdp, don't use the packaged Rust version on Ubuntu, but instead get Rust using https://rustup.rs/

greg7mdp:
Sorry, I'm a rust beginner.
I have the latest version installed, but when I use cargo inside `ron` directory it switches back to 1.36.
Here it is on windows:

```
C:\greg\github>where cargo
C:\Users\Valued Customer\.cargo\bin\cargo.exe

C:\greg\github>cargo --version
cargo 1.56.0 (4ed5d137b 2021-10-04)

C:\greg\github>cd ron

C:\greg\github\ron>cargo clean

C:\greg\github\ron>cargo --version
cargo 1.36.0 (c4fcfb725 2019-05-15)

```

and if I remove the 1.36 toolchain cargo reinstalls it:

```
C:\greg\github\ron>rustup toolchain uninstall 1.36.0-x86_64-pc-windows-msvc
info: uninstalling toolchain '1.36.0-x86_64-pc-windows-msvc'
info: toolchain '1.36.0-x86_64-pc-windows-msvc' uninstalled

C:\greg\github\ron>cargo clean
info: syncing channel updates for '1.36.0-x86_64-pc-windows-msvc'
info: latest update on 2019-07-04, rust version 1.36.0 (a53f9df32 2019-07-03)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
[...]

PR #340, opened 2021-11-17
chore(deps): use exact version for bitflags

torkleyy:
Motivation: allow building with MSRV without patching lock file - see #332

Usually this is discouraged, but `bitflags` only provides a macro that generates code for us, so if it gets included with multiple different versions in a dependency tree it won't cause any issues.

* ~~[x] I've included my change in `CHANGELOG.md`~~ not necesary

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,12 +18,9 @@ repository = "https://github.com/ron-rs/ron"
 documentation = "https://docs.rs/ron/"
 exclude = ["bors.toml", ".travis.yml"]
 
-[lib]
-name = "ron"
-
 [dependencies]
 base64 = "0.13"
-bitflags = "1.0.4"
+bitflags = "=1.2.1"
 indexmap = { version = "1.0.2", features = ["serde-1"], optional = true }
 serde = { version = "1.0.60", features = ["serde_derive"] }
 


torkleyy:
Good point, I wasn't aware `indexmap` is also an issue. However, forcing an exact version for it has more implications than it has for `bitflags`...

> Afterwards, the workaround in the CI could probably also be removed.

Yes, I'll do that :+1:

MomoLangenstein:
> 
> 
> Good point, I wasn't aware `indexmap` is also an issue. However, forcing an exact version for it has more implications than it has for `bitflags`...

That is true ... it seems that `hashbrown` is the problem there as it uses the `ptr_cast` feature that wasn't stable yet in `1.36.0`

torkleyy:
I'll merge this as-is, it's already a bit of an improvement.

Imberflur:
> `bitflags` only provides a macro that generates code for us, so if it gets included with multiple different versions in a dependency tree it won't cause any issues.

IME cargo will fail to compile rather than including multiple compatible versions of a crate in the tree, so anyone depending on `ron` won't be able to use a newer version of `bitflags` (unless they release a new major version)

torkleyy:
@Imberflur that's only when you have that dependency in your public API.

Imberflur:
e.g. see https://github.com/nix-rust/nix/issues/1548

torkleyy:
Oh I see. My understanding was that Cargo can handle multiple versions of a dependency in the tree, which is the case, but with the limitation that all versions must be semver incompatible, which isn't true for bitflags...

PR #341, opened 2021-11-19
Fix #338 value map roundtrip

### LINKED ISSUE #338: Round-trip failure: Empty maps/sequences gain an extra layer of nesting
This example program probably ought to exit successfully
```rust
fn main() {
    let input = "{}";
    let v: ron::Value = ron::from_str(input).unwrap();
    println!("{:?}", v);
    let ser = ron::to_string(&v).unwrap();
    println!("{:?}", ser);
    let roundtrip = ron::from_str(&ser).unwrap();
    println!("{:?}", roundtrip);
    assert_eq!(v, roundtrip);
}
```
But it doesn't:
```
Map(Map({}))
"({})"
Seq([Map(Map({}))])
  left: `Map(Map({}))`,
 right: `Seq([Map(Map({}))])`', src/main.rs:9:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This actually continues indefinitely. `"{}"` becomes `"({})"` which becomes `"[({})]"` and from then on we just add square brackets every round-trip.

MomoLangenstein:
Fixes #338. Sometimes it's as simple as that :)


torkleyy (approved):
:shipit:

Issue #301, opened 2021-04-17
Unhelpful error message when struct name doesn't match


cswinter:
When deserializing a string that defines a RON struct with an optional class name that does not match the name of the Rust struct, `ron::from_str` returns an error that displays as `Error: 1:1: Expected struct`. It took me quite a long time to figure out what the issue was. It would be more helpful if the error message was something like `Error: 1:1: Expected struct "Name" but got struct "ActualName"`.

Valmirius:
I'm also getting this error "Failed to load config: 1:1: Expected struct" in the decode_file.rs example via example.ron even when using the exact same files, not entirely sure why it's asking for a Struct.

MomoLangenstein:
I just had a very brief look at the codebase to see how difficult it would be to implement this. Here are my thoughts:

1) A simple solution would be to also report the name of the struct that was expected. This would already give most of the debugging value as you could then manually look at the RON file to see what name you provided. Implementing this option is trivial.

2) To find out if a different identifier was given instead, `ron` would have to check if any valid identifier but wrong identifier is given and that the opening `(` exists. In the case where the only error is the wrong struct name, this could certainly be achieved, though at the cost of a bit of extra code clutter. However, any further edge cases would still give the original error messages.

Personally, I think that option (1) could be a good start as it provides most of the error reporting benefit already while requiring very few code changes and no additional code complexity. Option (2) could then still be implemented later on as a refinement.

PR #342, opened 2021-11-19
Added improved `ExpectedStruct` error message

MomoLangenstein:
Improves the error messages for `ExpectedStruct` (now `ExpectedNamedStruct`), `NoSuchExtension`, and `UnexpectedByte` by including the erroneous identifier in the message. This small change should help with #301.


MomoLangenstein:
@torkleyy In relation to https://github.com/ron-rs/ron/issues/301#issuecomment-973997228 I did some small experimentation to ever so slightly improve the error message from `Expected struct` to `Expected struct 'MyStruct'` when the struct name is known. This would already help the case when an error message is confusing but doesn't require the complex changes that would be needed to see e.g. if a different struct name was given.

src/de/mod.rs
```diff
@@ -409,8 +409,10 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             } else {
                 self.bytes.err(ErrorCode::ExpectedStructEnd)
             }
-        } else {
+        } else if name.is_empty() {
             self.bytes.err(ErrorCode::ExpectedStruct)
```
torkleyy, line 413: Are there cases where `name` is empty?

torkleyy (commented):
Great quality of life improvement! Can we always include the struct name in the error?

src/de/mod.rs
MomoLangenstein: For newtype structs this would only be the case if a serde impl gives us an empty name. For structs it can actually happen when a struct variant is deserialised, as I did not find a way to get the name of the variant. I just added the same check here for consistency.

torkleyy (approved):
Alright, LGTM!

MomoLangenstein:
> Alright, LGTM!

Ok, I'll add the Changelog entry and merge it later today :)

MomoLangenstein:
@torkleyy Ready to merge?

torkleyy (approved):
:shipit:

PR #343, opened 2021-11-19
Expose `Extensions` during ser+de through `ron::Options`

MomoLangenstein:
This PR is based on #281 and #339 (with a potential application for #334 as well).

1) It adds a new `Options` struct from which all ser+de utils methods can be called.
2) The existing utils methods `XXX` now use `Options::default().XXX` as their implementation.
3) The `Options` struct can be configured with a set of default RON extensions
  a) During deserialization default extensions are automatically enabled, i.e. do not have to be included in the RON.
  b) During serialization default extensions are automatically enabled and are **not** included in the output RON. Additional extensions enabled through the `PrettyConfig` which are not in the set of default extensions are still included in the output RON. Therefore, this PR supersedes #339.
  c) This now enables programs which always use a set of extensions to enable them programmatically and exclude them from the RON config files.
4) This PR is fully backwards compatible by adding one extra method each to the `Serializer` and `Deserializer` and not changing the default behaviour (i.e. no default extensions).
5) This PR also does a tiny bit of cleanup in the docs and `Default` impl for `PrettyConfig`.

* [x] Add and update the documentation

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -35,12 +35,16 @@ pub struct Options {
     ///  activation is NOT included in the output RON.
     /// No extensions are enabled by default.
     pub default_extensions: Extensions,
+    /// Private field to ensure adding a field is non-breaking.
+    #[serde(skip)]
+    _future_proof: (),
 }
 
 impl Default for Options {
     fn default() -> Self {
         Self {
             default_extensions: Extensions::empty(),
+            _future_proof: (),
         }
     }
 }
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,31 +1,60 @@
+//! Roundtrip serde Options module.
+
 use std::io;
 
-use serde::{de, ser};
+use serde::{de, ser, Deserialize, Serialize};
 
 use crate::de::Deserializer;
 use crate::error::Result;
 use crate::extensions::Extensions;
 use crate::ser::{PrettyConfig, Serializer};
 
+/// Roundtrip serde options.
+///
+/// # Examples
+///
+/// ```
+/// use ron::{Options, extensions::Extensions};
+///
+/// let ron = Options::default()
+///     .with_default_extension(Extensions::IMPLICIT_SOME);
+///
+/// let de: Option<i32> = ron.from_str("42").unwrap();
+/// let ser = ron.to_string(&de).unwrap();
+///
+/// assert_eq!(ser, "42");
+/// ```
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(default)]
 pub struct Options {
-    default_extensions: Extensions,
+    /// Extensions that are enabled by default during serialization and
+    ///  deserialization.
+    /// During serialization, these extensions do NOT have to be explicitly
+    ///  enabled in the parsed RON.
+    /// During deserialization, these extensions are used, but their explicit
+    ///  activation is NOT included in the output RON.
+    /// No extensions are enabled by default.
+    pub default_extensions: Extensions,
 }
 
-impl Options {
-    #[must_use]
-    pub fn build() -> Self {
+impl Default for Options {
+    fn default() -> Self {
         Self {
             default_extensions: Extensions::empty(),
         }
     }
+}
 
+impl Options {
     #[must_use]
+    /// Enable `default_extension` by default during serialization and deserialization.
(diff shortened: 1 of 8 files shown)


MomoLangenstein:
@torkleyy Here's my initial attempt at the `Option` - I'm looking forward to your feedback!

src/options.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[must_use]
+    pub fn build() -> Self {
+        Self {
+            default_extensions: Extensions::empty(),
+        }
+    }
+
+    #[must_use]
+    pub fn with_default_extensions(mut self, default_extensions: Extensions) -> Self {
+        self.default_extensions = default_extensions;
```
torkleyy, line 24:
```suggestion
        self.default_extensions |= default_extensions;
```

torkleyy (commented):
Sorry for the late review. I like where this is going!

src/options.rs
MomoLangenstein: That would definitely allow for nicer code - do you think it would then also need a separate replacement / reset method?

torkleyy: I'm not sure why that would be needed. Unless you got a specific case in mind, I wouldn't add that.

MomoLangenstein: Ok, thanks for your assessment. I didn't have anything in mind really, my usecase has a constant config that I only need to build once.

torkleyy: If the extensions are just assigned, there's no need to make use of the builder pattern imo - I guess we could offer both, a builder and a "constructor".

MomoLangenstein: I agree that right now a builder is certainly overkill. But maybe additional config options will be added in the future, when having a constructor with lots of parameters would become annoying.

MomoLangenstein:
> Sorry for the late review. I like where this is going!

@torkleyy If you like the API as it is right now (and you give me the go-ahead), I'll start updating the documentation.

torkleyy (approved):
Looks great, go ahead :+1:

MomoLangenstein:
@torkleyy I've added documentation and the CHANGELOG entry. I also made the following changes to the `Options` API:
* `Options::build()` is now `Options::default()`
* both `Serializer` and `Deserializer` now take the `Options` in their new constructors instead of the `default_extensions`. This makes `Options` more obtrusive but makes it much easier to add new options to the `Options`, e.g. for #334

I've also made some minor code cleanup adjustments:
* moved the `Position` from `parse` to `errors` as it is exposed anyways (both in `de` and `error` and rustdoc was not showing them as the exact same struct)
* made `ser::State` private
* hid the doc for `ser::Compound` as it is an implementation detail but cannot be private
* simplified `PrettyConfig`'s `Default` impl so that it isn't duplicated between serde and `Default`

src/ser/mod.rs
```diff
@@ -77,25 +70,18 @@ struct Pretty {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct PrettyConfig {
     /// Limit the pretty-ness up to the given depth.
-    #[serde(default = "default_depth_limit")]
```
torkleyy, line 80: I'm not sure how this is meant to work, if you remove these serde attributes it will default to 0 or empty strings.

src/options.rs
```diff
@@ -0,0 +1,146 @@
...
+///     .with_default_extension(Extensions::IMPLICIT_SOME);
+///
+/// let de: Option<i32> = ron.from_str("42").unwrap();
+/// let ser = ron.to_string(&de).unwrap();
+///
+/// assert_eq!(ser, "42");
+/// ```
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(default)]
+pub struct Options {
```
torkleyy, line 29: We could add `non_exhaustive` for this struct.

MomoLangenstein: Thank you for that suggestion! I've used the same approach as in `PrettyConfig` to fix this

src/ser/mod.rs
MomoLangenstein: You are absolutely right, I forgot to add `#[serde(default)]` (https://serde.rs/container-attrs.html#default) to the struct. Should be fixed now

torkleyy: Oh, interesting, did not know about the container-level default attribute.

torkleyy (approved):

Issue #290, opened 2020-12-06
No space after semicolons & comma after struct fields is not "pretty"


Boscop:
Consider this:
```rust
        [
            Some((
                asset: Misc(Buffer),
                sampler: (filter:mipmap,wrap:clamp,vflip:true), // should have spaces after semicolons & commas
            )),// [0]
            None,// [1]
            None,// [2]
            Some((
                asset: Misc(Stream),
                sampler: (filter:linear,wrap:clamp,vflip:true),
            )),
        ],
```

https://docs.rs/ron/0.6.2/ron/ser/struct.PrettyConfig.html#structfield.depth_limit

The types after this depth should still follow normal formatting guidelines like having spaces after semicolons & commas.

> depth_limit: usize
> Limit the pretty-ness up to the given depth.

What does this even mean? That everything after that level is not pretty-printed?
I use `.with_depth_limit(4)` but I don't want everything beyond that depth to not be pretty,
I just want to limit the depth for expanding types to multiple lines.

Actually what I'd **really** want is some `max_width` limit like in rustfmt to say that types should be written on one line unless they go over that limit. I thought that's what the depth limit is for, but apprently not.
Can you please add something like `max_width`? :)

kvark:
> What does this even mean? 

I think having spaces after commas would still be nice there. But the idea was to stop breaking the lines, at least.

`max_width` is not something I needed personally, but if you want to contribute this, I can believe it's going to be useful to some.

kvark:
comment

MomoLangenstein:
@Boscop Have #344 and #349 solved your issue?

PR #349, opened 2021-12-18
Add compact arrays (#299) and pretty inline separators

MomoLangenstein:
This PR includes the implementation for compact arrays from #299 (stale and quite tricky to rebase), and a fix for #290 to make the pretty RON prettier. Specifically, this means that:
- pretty tuples and arrays will now use a separator string after `,` when multiline is disabled
- pretty structs and maps will now use a separator string after `:` and `,`
- this separator string is a single space by default

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -76,6 +76,8 @@ pub struct PrettyConfig {
     pub new_line: String,
     /// Indentation string
     pub indentor: String,
+    /// Separator string
+    pub separator: String,
     // Whether to emit struct names
     pub struct_names: bool,
     /// Separate tuple members with indentation
@@ -87,6 +89,8 @@ pub struct PrettyConfig {
     /// Enable extensions. Only configures 'implicit_some',
     ///  'unwrap_newtypes', and 'unwrap_variant_newtypes' for now.
     pub extensions: Extensions,
+    /// Enable compact arrays
+    pub compact_arrays: bool,
     /// Private field to ensure adding a field is non-breaking.
     #[serde(skip)]
     _future_proof: (),
@@ -124,6 +128,15 @@ impl PrettyConfig {
     /// Default: 4 spaces
     pub fn indentor(mut self, indentor: String) -> Self {
         self.indentor = indentor;
+
+        self
+    }
+
+    /// Configures the string sequence used to separate items inline.
+    ///
+    /// Default: 1 space
+    pub fn separator(mut self, separator: String) -> Self {
+        self.separator = separator;
 
         self
     }
@@ -170,6 +183,23 @@ impl PrettyConfig {
         self
     }
 
+    /// Configures whether every array should be a single line (true) or a multi line one (false)
+    /// When false, `["a","b"]` (as well as any array) will serialize to
+    /// `
+    /// [
+    ///   "a",
+    ///   "b",
+    /// ]
+    /// `
+    /// When true, `["a","b"]` (as well as any array) will serialize to `["a","b"]`
+    ///
+    /// Default: `false`
+    pub fn compact_arrays(mut self, compact_arrays: bool) -> Self {
+        self.compact_arrays = compact_arrays;
+
+        self
+    }
+
     /// Configures extensions
     ///
     /// Default: Extensions::empty()
@@ -190,11 +220,13 @@ impl Default for PrettyConfig {
                 String::from("\r\n")
             },
             indentor: String::from("    "),
+            separator: String::from(" "),
             struct_names: false,
             separate_tuple_members: false,
             enumerate_arrays: false,
             extensions: Extensions::empty(),
             decimal_floats: false,
+            compact_arrays: false,
             _future_proof: (),
         }
     }
@@ -263,13 +295,6 @@ impl<W: io::Write> Serializer<W> {
         })
     }
 
-    fn is_pretty(&self) -> bool {
-        match self.pretty {
-            Some((ref config, ref pretty)) => pretty.indent <= config.depth_limit,
-            None => false,
-        }
-    }
-
     fn separate_tuple_members(&self) -> bool {
         self.pretty
             .as_ref()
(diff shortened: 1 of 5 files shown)


torkleyy (commented):
Is it useful in practice to turn compact arrays completely on or off? It seems to me I'd want arrays with few and short elements compact, but as soon as the elements are non-primitive I'd probably want them all in their own line.
However, handling all the cases seems pretty hard as well...

src/ser/mod.rs
```diff
@@ -291,12 +316,16 @@ impl<W: io::Write> Serializer<W> {
     }
 
     fn start_indent(&mut self) -> Result<()> {
+        self.start_indent_if(|_| true)
+    }
+
+    fn start_indent_if(&mut self, condition: impl Fn(&PrettyConfig) -> bool) -> Result<()> {
```
torkleyy, line 322: Why use a closure here if it only receives the config which is known ahead of time?
Maybe calling it `expand` or something would make the code more self-explanatory.

MomoLangenstein:
> Is it useful in practice to turn compact arrays completely on or off? It seems to me I'd want arrays with few and short elements compact, but as soon as the elements are non-primitive I'd probably want them all in their own line. However, handling all the cases seems pretty hard as well...

I think the same question came up in the original PR. I think the best we could do is check the optionally sequence length against some threshold for finer control. While doing things based on type info would be awesome, I have no idea how this could be done in serde.

torkleyy:
I don't think we can do that ahead of time. Also, I don't think we should try to cover all the possible formatting of RON. A full-blown formatter inside of the serialization code just makes is it unreadable and bloated - and I don't think other serialization libraries offer as much configuration as we have right now.

torkleyy:
So my proposal:

1. Decide what's useful enough to have inside of RON
2. More involved formatting should be implemented in an external formatter

src/ser/mod.rs
MomoLangenstein: I'm not sure what you want to rename to `expand`. The closure usage comes from the original PR ... I think the extra helper functions could be avoided since they are all just used once and enclosing them inside an if at the call site should work as well.

torkleyy: You're right, that's even better!

MomoLangenstein:
> So my proposal:
>     1. Decide what's useful enough to have inside of RON
>     2. More involved formatting should be implemented in an external formatter

That makes sense ... I assume this formatter would perform some light parsing of RON and take in a subset of the options `PrettyConfig` handles right now and transform the `String`?

torkleyy:
> > So my proposal:
> > 1. Decide what's useful enough to have inside of RON
> > 2. More involved formatting should be implemented in an external formatter
> 
> That makes sense ... I assume this formatter would perform some light parsing of RON and take in a subset of the options `PrettyConfig` handles right now and transform the `String`?

Do you mean a superset? And yes, it will have to do its own parsing because

* we need look-ahead to determine whether we want to format in expanded or compact form
* parsing into something like `Value` and then formatting that would discard the comments

MomoLangenstein:
> > > So my proposal:
> > > 
> > > 1. Decide what's useful enough to have inside of RON
> > > 2. More involved formatting should be implemented in an external formatter
> > 
> > 
> > That makes sense ... I assume this formatter would perform some light parsing of RON and take in a subset of the options `PrettyConfig` handles right now and transform the `String`?
> 
> Do you mean a superset?

Maybe a superset of a subset? I think the options `struct_names`, `decimal_floats`, and `extensions` would all remain as they can actually impact the RON besides some nicer whitespace usage.

> And yes, it will have to do its own parsing because
> 
>     * we need look-ahead to determine whether we want to format in expanded or compact form
> 
>     * parsing into something like `Value` and then formatting that would discard the comments

What if this new `Value` also was a `Serializer` so that if you want pretty RON you could go into it directly without serializing to a `String` and then parsing the the `String` again for formatting?

torkleyy:
> Maybe a superset of a subset? I think the options `struct_names`, `decimal_floats`, and `extensions` would all remain as they can actually impact the RON besides some nicer whitespace usage.

Ah, yes, that makes sense!

> What if this new `Value` also was a `Serializer` so that if you want pretty RON you could go into it directly without serializing to a `String` and then parsing the the `String` again for formatting?

Yes, certainly, I think it should be. I've even started such an attempt here: https://github.com/ron-rs/ron-reboot
So it's not a new `Value` type but really an `AST` that you can deserialize and serialize. `ron-reboot` works as it is, but some problems remain (which is also why I didn't get the word out):

* I don't really like the name. I don't think it should replace this `ron` repository but rather extend it
* The performance isn't acceptable. I knew it was gonna be slower due to intermediate representations, but it seems way too slow to me - I suspect a bug where it scans the whole input or something like that

src/ser/mod.rs
MomoLangenstein: This should be better now. The old version also had a formatting bug, which is why I added an extra test - in compact arrays the indentation level should *not* be incremented.

MomoLangenstein:
> Yes, certainly, I think it should be. I've even started such an attempt here: https://github.com/ron-rs/ron-reboot So it's not a new `Value` type but really an `AST` that you can deserialize and serialize. `ron-reboot` works as it is, but some problems remain (which is also why I didn't get the word out):
> 
>     * I don't really like the name. I don't think it should replace this `ron` repository but rather extend it
> 
>     * The performance isn't acceptable. I knew it was gonna be slower due to intermediate representations, but it seems way too slow to me - I suspect a bug where it scans the whole input or something like that

That looks really cool!

MomoLangenstein:
So what's the plan now? Should this PR get merged and at some point the options move to your formatter?

torkleyy:
Yes, I think we should merge this but not accept any further configuration in the future, because I fear with every option the code becomes harder to maintain.
I'm hoping we can make such a formatter in the future but obviously it will require some work... I'm hoping I can figure out what's causing the performance problems some time next year and then continue from there and implement a basic `ronfmt`.

MomoLangenstein:
> Yes, I think we should merge this but not accept any further configuration in the future, because I fear with every option the code becomes harder to maintain. I'm hoping we can make such a formatter in the future but obviously it will require some work... Maybe I can figure out what's causing the performance problems some time next year and then continue from there and implement a basic `ronfmt`.

Ok, that's sounds like a good plan!

torkleyy (approved):

PR #351, opened 2021-12-18
Guard `u128` and `i128` behind optional `integer128` feature

MomoLangenstein:
This PR adds an optional `integer128` feature that guards the usage of `i128` and `u128`. Therefore, any library using `serde` with either of these types should now enable the feature. This PR is heavily inspired by #304.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,10 @@ repository = "https://github.com/ron-rs/ron"
 documentation = "https://docs.rs/ron/"
 exclude = ["bors.toml", ".travis.yml"]
 
+[features]
+default = ["integer128"]
+integer128 = []
+
 [dependencies]
 base64 = "0.13"
 bitflags = "=1.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ documentation = "https://docs.rs/ron/"
 exclude = ["bors.toml", ".travis.yml"]
 
 [features]
-default = ["integer128"]
+default = []
 integer128 = []
 
 [dependencies]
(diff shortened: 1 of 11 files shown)


MomoLangenstein:
I've started the implementation to add an `integer128` feature. So far the functionality should be there, but some aspects still have to be refactored (e.g. the serialize dispatch code duplication). What do you think of the general design? How should the CI be adapted? I know we can check if compilation for wasm succeeds, but should e.g. the tests be run as well?

Cargo.toml
```diff
@@ -18,6 +18,10 @@ repository = "https://github.com/ron-rs/ron"
 documentation = "https://docs.rs/ron/"
 exclude = ["bors.toml", ".travis.yml"]
 
+[features]
+default = ["integer128"]
```
kvark, line 22: Is there a good reason to have it on by default?
I believe generally having less default features is better, unless a feature is *really* expected.

src/parse.rs
```diff
@@ -110,6 +112,15 @@ pub struct Bytes<'a> {
     line: usize,
 }
 
+#[cfg(feature = "integer128")]
+pub(crate) type UInt = u128;
```
kvark, line 116: Could we perhaps call these typedefs differently? Like `LargeUint` or something to indicate they aren't regular integers

kvark (commented):
Thank you!
I believe the direction is sound

Cargo.toml
MomoLangenstein: While I do expect serde to work for 128b integers, I do agree that non-default features are easier to handle for the end-user.

src/parse.rs
MomoLangenstein: Done

torkleyy (approved):
Looks good to me!

kvark (approved):

Issue #152, opened 2019-03-15
Edge case where RON fails but all other serde format types succeed


jaynus:
Using any format of bitflags serialization, I seem to have found, somehow, a way in which RON fails to deserialize values, but all other serde serializers seem to handle just fine. I don't understand serde enough to understand what is incorrect about the RON implementation.

Error is: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Message("invalid type: string \"One\", expected a borrowed string")', src/libcore/result.rs:997:5

Specifically, I could at least trace the error down to this line: https://github.com/kvark/bitflags-serial/blob/master/src/lib.rs#L66
or this line
https://github.com/H2CO3/option_set/blob/master/src/lib.rs#L242

I've created a repo minimally showing the issue here: https://github.com/jaynus/ron_fail_poc

I ran into this using RON originally with bitflags_serial (https://github.com/kvark/bitflags-serial). I then swapped to option_set in an attempt to fix it, and ran into the same issue.

kvark:
Any idea how does this differ from a general case that is expected to work with RON?

jaynus:
How do you mean exactly? I've updated the test case showing working cases vs. non-working cases with bitflags vs. option_set. The update shows that bitflags works for both ron and serde_json, while option_set works with serde_json but NOT ron.

I'm not sure what details you need.

Update here:
https://github.com/jaynus/ron_fail_poc/commit/fe5bb3d4f35640b8bec40dcce30ecf6d2354cdb5

kvark:
Hm weird. I mean, the `bitflags-serial` was specifically implemented with Ron in mind and tested on it, see https://github.com/bitflags/bitflags/issues/147
So the fact it no longer works means there was a regression in one of the pieces.

jaynus:
Oh! I totally failed to connect the dots that that was you! I've added bitflags_serial to the test case.

I've added a test case for bitflags_serial, which also appears to fail for JSON and RON.

https://github.com/jaynus/ron_fail_poc/commit/faa4ce57bee113176a44b397371976fdea4fcdf6

jaynus:
As an additional, I've noticed there are cases where it DOES work with amethyst assets RonFormat loader, though. 

So it appears maybe its a difference between from_str and from_bytes? I cannot replicate from_bytes working though in my case.

jaynus:
Update: I've also confirmed it fails with stable (I am using latest nightly).

jaynus:
[...]

PR #352, opened 2021-12-25
Add test to check that #152 is fixed

MomoLangenstein:
Adds an extra unit test to make sure that #152 remains fixed.


MomoLangenstein:
@kvark Since the originally given failing case given by @jaynus in https://github.com/ron-rs/ron/issues/152#issue-421298302 tests `bitflags`, `option_set`, and `bitflags-serial`, I have added the latter two as dev-dependencies. If you'd prefer, the test could also be reduced to just testing `bitflags` without adding new dev-dependencies.

kvark:
Thank you!
I think it's useful to keep bitflags-serial checked, given that there is still no upstream solution to replace it.

kvark (approved):

PR #353, opened 2021-12-25
Fix #265 with better missing comma detection

### LINKED ISSUE #265: Missing commas produce unhelpful errors with col/line 0
This code:
```rust
    #[derive(serde::Deserialize)]
    struct Test {
        a: i32,
        b: i32,
    }
    let string = r#"Test(
        a: 1 // <-- forgetten comma here
        b: 2
    )"#;

    ron::from_str::<Test>(string).unwrap();
```
Produces the following error:
```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: Message("missing field `b`"), position: Position { col: 0, line: 0 } }', src/main.rs:67:35
```

Which is really hard to understand because 
1. it says field `b` is missing but it doesn't
2. it doesn't say anything about commas
3. it points to `0, 0`

MomoLangenstein:
Fix #265 by better dealing with missing comma in `CommaSeparated` so that an `ExpectedComma` error can be reported.


MomoLangenstein:
?r @torkleyy This should be a quick fix as it really just detects the error earlier without needing any extra information

torkleyy (approved):
Awesome!

PR #354, opened 2021-12-26
Add struct name mismatch error + better error messages

### LINKED ISSUE #301: Unhelpful error message when struct name doesn't match

MomoLangenstein:
Fixes #301 by adding a new `ExpectedStructName` error variant when a struct name is found but doesn't match the expected one. This PR also improves the error messages to be clearer about what is expected in the RON string, not how it is implemented (e.g. `Expected opening (` instead of `Expected struct`)


MomoLangenstein:
?r @torkleyy

torkleyy (approved):
:shipit:

PR #356, opened 2021-12-30
Centralise error positioning + cover custom de errors

### LINKED ISSUE #203: Some errors are missing line number information.
Hi,
I'm not very familiar with serde so I don't know if this is a problem with that, but certain error messages are pretty unhelpful right now for fixing errors. In particular, a great deal of errors are missing information about what line they came from (or just have it baked into the message - this is better than nothing, but it would be more pleasant to actually get the location out so I can present it to the user.)

As an example step, the following code (edited from the example):
```rust
use ron::de::from_str;
use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct Config {
    boolean: bool,
    tuple: (u32, u32),
}

const CONFIG: &str = "
(
    boolean: true, 
    //tuple: (3 /*(2 + 1)*/, 7 /*(2 * 5 - 3)*/), //Notice the commented out field here
)";

fn main() {
    let config: Config = match from_str(CONFIG) {
        Ok(x) => x,
        Err(e) => {
            panic!("{:?}", e)
        }
    };

    println!("Config: {:?}", &config);
}
```
will just print out `missing field \`tuple\``, which isn't super helpful in larger files.

I took a look around in the code to see if there was anything I could hack in myself to help with this, but I had a hard time actually tracing the path the code follows.

MomoLangenstein:
Fixes #203 and supersedes #209. Also includes some small minor refactoring of the `ErrorCode` names.

In particular, the `Deserializer` now uses `ErrorCode` as its error type, no longer having to pretend to produce error positions in all cases (which as #209 showed were missing in all serde-internal and user-generated cases, and can only be added during deserialization to the former). Instead, the exposed helper functions `from_str` etc. now add the positional information afterwards, which cleans up the code and gives us positions for custom errors in serde-code and user-code for free (though the position still only points at the character after the error or directly at it).

### DIFF
--- a/src/de/id.rs
+++ b/src/de/id.rs
@@ -1,6 +1,8 @@
 use serde::de::{self, Visitor};
 
-use super::{Deserializer, Error, Result};
+use super::{Deserializer, ErrorCode};
+
+type Result<T> = std::result::Result<T, ErrorCode>;
 
 pub struct IdDeserializer<'a, 'b: 'a> {
     d: &'a mut Deserializer<'b>,
@@ -13,7 +15,7 @@ impl<'a, 'b: 'a> IdDeserializer<'a, 'b> {
 }
 
 impl<'a, 'b: 'a, 'c> de::Deserializer<'b> for &'c mut IdDeserializer<'a, 'b> {
-    type Error = Error;
+    type Error = ErrorCode;
 
     fn deserialize_identifier<V>(self, visitor: V) -> Result<V::Value>
     where
--- a/src/de/id.rs
+++ b/src/de/id.rs
@@ -1,8 +1,6 @@
 use serde::de::{self, Visitor};
 
-use super::{Deserializer, ErrorCode};
-
-type Result<T> = std::result::Result<T, ErrorCode>;
+use super::{Deserializer, Error, Result};
 
 pub struct IdDeserializer<'a, 'b: 'a> {
     d: &'a mut Deserializer<'b>,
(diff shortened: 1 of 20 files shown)


MomoLangenstein:
?r @torkleyy @kvark

MomoLangenstein:
There is also the question whether `ron` should split the error type into two, one for `ser` and one for `de`, or keep the combined one (like `serde_json`). What are your thoughts? For `ser` it seems like we only generate `Io` and `Message` errors that are never positioned anyways (unless we wanted to arbitrarily track a fake line and column somehow).

MomoLangenstein:
?r @torkleyy Any thoughts?

kvark:
I'm very sorry that neither me or torkley are responsive!
If you are confident with the move, please feel free to commit without reviews if N days passed without any feedback (as long as this deadline is visible). For this PR, I'll review it now.

src/de/id.rs
```diff
@@ -1,6 +1,8 @@
 use serde::de::{self, Visitor};
 
-use super::{Deserializer, Error, Result};
+use super::{Deserializer, ErrorCode};
+
+type Result<T> = std::result::Result<T, ErrorCode>;
```
kvark, line 5: perhaps another typedef for `Result<T, Error>` would help here?

kvark (approved):

src/de/id.rs
MomoLangenstein: I've added a second typedef, now for the naming. I've remapped `ErrorCode` => `Error` as this is now the error type returned by both the `Serializer` and `Deserializer`, `type Result<T, E = Error> = std::result::Result<T, E>` defaults to this error type as well but allows importing the `ron` `Result` without losing the functionality of `std::result::Result`. Furthermore, `Error` => `SpannedError` and `type SpannedResult<T> = std::result::Result<T, SpannedError>`, which is returned by all convenience function around deserialization. I'm not sure if `SpannedError` is the best name, `PositionalError` could also work. This renaming (well, this PR in general) will break some code at the benefit of clarifying that only deserialization errors can have nice error position information + they will now always have this.

kvark (approved):

PR #363, opened 2022-01-28
Enable `decimal_floats` by default for `Value::Number(Float)` roundtrip

### LINKED ISSUE #337: Round-trip failure: ser -> de transforms Float(1.0) into Integer(1)
This example program probably ought to exit successfully
```rust
fn main() {
    let input = "1.0";
    let v: ron::Value = ron::from_str(input).unwrap();
    println!("{:?}", v);
    let ser = ron::to_string(&v).unwrap();
    println!("{:?}", ser);
    let roundtrip = ron::from_str(&ser).unwrap();
    println!("{:?}", roundtrip);
    assert_eq!(v, roundtrip);
}
```
But it doesn't:
```
Number(Float(Float(1.0)))
"1"
Number(Integer(1))
  left: `Number(Float(Float(1.0)))`,
 right: `Number(Integer(1))`', src/main.rs:9:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This seems plausible as a ser/de behavior if RON only has a number type, but if that's the case `ron::Value::Number` should depend only on the numeric value not the discriminant.

MomoLangenstein:

Fixes #337

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix issue [#152](https://github.com/ron-rs/ron/issues/152) bitflags serde ([#352](https://github.com/ron-rs/ron/pull/352))
 - Fix issue [#359](https://github.com/ron-rs/ron/issues/359) with `DeserializeSeed` support ([#360](https://github.com/ron-rs/ron/pull/360))
 - Bump MSRV to 1.46.0 ([#361](https://github.com/ron-rs/ron/pull/361))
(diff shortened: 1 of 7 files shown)


MomoLangenstein:
?r @torkleyy This PR simply makes `decimal_floats` on by default
(a) as it fixes #337 as integer floats are now serialized as decimals by default
(b) imo it also looks nicer

kvark (approved):
Does it even need to be an option?
I would think just removing the option entirely and doing this unconditionally would be better.

MomoLangenstein:
> @kvark  Does it even need to be an option? I would think just removing the option entirely and doing this unconditionally would be better.

That's a good idea - thanks for the suggestion! This commit will of course require a version bump as it is a breaking change, though I guess the usefulness of not including a decimal was questionable from the start.

PR #366, opened 2022-02-20
Add code coverage CI step

### LINKED ISSUE #176: Add code coverage badge
We should measure code coverage by CI and add a badge to the repo, goal being to increase have RON fully covered by tests.

MomoLangenstein:
Fixes #176

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -73,12 +73,10 @@ jobs:
       - run: cargo install grcov --force --locked
       - run: |
           cargo test --all-features --all-targets
-          cargo run --example decode_file
-          cargo run --example decode
-          cargo run --example encode
-          cargo run --example transcode
           grcov . -s . --binary-path ./target/debug/ \
             -t cobertura -o cobertura.xml --branch \
+            --keep-only "src/*" \
+            --keep-only "tests/*" \
             --ignore-not-existing \
             --excl-line GRCOV_EXCL_LINE \
             --excl-start GRCOV_EXCL_START \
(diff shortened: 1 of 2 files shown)


MomoLangenstein:
@kvark @torkleyy I'm hoping that just adding the CI step and badge should just work. Ideally, we wouldn't need to reinstall `grcov` on every CI run, but the [actions-rs/install](https://github.com/actions-rs/install) action that could solve this still has security issues - potentially this might become an incremental improvement later on. I've added `GRCOV_EXCL_LINE`,  `GRCOV_EXCL_START`, and `GRCOV_EXCL_STOP` markers so we can exclude sections from coverage analysis.

MomoLangenstein:
If we don't find any issues with this PR, I'll merge it by February 28th.

.github/workflows/ci.yaml
```diff
@@ -58,3 +58,35 @@ jobs:
...
+        with:
+          toolchain: nightly
+          profile: minimal
+          components: llvm-tools-preview
+          override: true
+      - run: cargo install grcov --force --locked
+      - run: |
+          cargo test --all-features --all-targets
+          cargo run --example decode_file
+          cargo run --example decode
```
torkleyy, line 77: I don't think examples should contribute to coverage.

torkleyy (approved):
Looks good otherwise!

.github/workflows/ci.yaml
MomoLangenstein: Ok, I removed the examples from the coverage runs and output - so they just won't be recorded at all and will also not drag down the coverage

torkleyy (approved):
:shipit:

PR #371, opened 2022-02-27
Fix #370 with better float EBNF and +inf -NaN +NaN parsing

### LINKED ISSUE #370: Inconsistencies in floating point parsing
Grammar taken from [current master](https://github.com/ron-rs/ron/blob/3ec0dc09f9b557d73790b9916b0d5cc4cc0f4c3d/docs/grammar.md). There are a couple inconcistencies between the grammar, the current implementation, and from how rust does it.

1. Exponential notation. The grammar allows `1.0e1`, but disallows `1e1` and bizzarely disallows sign: `1.0e+1` and `1.0e-1`. But the implementation accepts all those forms, and so does rust.

2. Underscores are disallowed in floats, and this is consistent in both the grammar and the implementation; but it differs from rust and is a bit unexpected. 

Here's small repro for everything:
```rust
type R<T> = ron::Result<T>;

fn main() {
    let x1: R<i64> = ron::from_str("1_0");
    println!("{:?}", &x1);
    // >>> Ok(10)

    let x2: R<f64> = ron::from_str("1_0.1_0");
    println!("{:?}", &x2);
    // >>> Err(Error { code: TrailingCharacters, position: Position { line: 1, col: 2 } })

    let x3: R<f64> = ron::from_str("1_0.10");
    println!("{:?}", &x3);
    // >>> Err(Error { code: TrailingCharacters, position: Position { line: 1, col: 2 } })

    let x4: R<f64> = ron::from_str("10.1_0");
    println!("{:?}", &x4);
    // >>> Err(Error { code: TrailingCharacters, position: Position { line: 1, col: 5 } })

    let x5: R<f64> = ron::from_str("1.0e-1");
    println!("{:?}", &x5);
    // >>> Ok(10.1)

    let y1: R<f64> = ron::from_str("1.0e1");
    println!("{:?}", &y1);
    // >>> Ok(10.0)

    let y2: R<f64> = ron::from_str("1e1");
    println!("{:?}", &y2);
    // >>> Ok(10.0)

    let y3: R<f64> = ron::from_str("1.0e1.0");
    println!("{:?}", &y3);
    // >>> Err(Error { code: ExpectedFloat, position: Position { line: 1, col: 1 } })

    let y4: R<f64> = ron::from_str("1.0e+1");
    println!("{:?}", &y4);
    // >>> Ok(10.0)

    let y5: R<f64> = ron::from_str("1.0e-1");
    println!("{:?}", &y5);
    // >>> Ok(0.1)
}
// y3 being Err is alright I believe, but the error message is weird
```

MomoLangenstein:
Fixes #370

Fixes the RON grammar for floats to be equivalent to `FromStr`'s grammar (which we use internally):
```ebnf
float = ["+" | "-"], ("inf" | "NaN" | float_num);
float_num = (float_int | float_std | float_frac), [float_exp];
float_int = digit, { digit };
float_std = digit, { digit }, ".", {digit};
float_frac = ".", digit, {digit};
float_exp = ("e" | "E"), ["+" | "-"], digit, {digit};
```

Consequently, RON now also accepts `+inf`, `-NaN`, and `+NaN`.

In comparison with `rustc`, `FromStr` still accepts floats without a leading digit, e.g. `.3` - these continue to work. The other way around, `rustc` accepts underscores in floats while `FromStr` and all other parsers I checked do not. I've added a special error `FloatUnderscore` to lint against this case.


MomoLangenstein:
@kvark @torkleyy I think this should be a simple fix - unless you specifically did not want to accept `+inf`, `-NaN`, and `+NaN`. Explicitly checking for the underscores in floats is of course a bit extra, but might help if someone wanted to copy-paste their Rust struct directly into a RON config.

kvark (approved):
Love those tests!

PR #372, opened 2022-03-30
Fixed serialization for floating point values less than EPSILON.

### LINKED ISSUE #374: Serialization error on very small floating point values
Very small floating point values serialize with an extra ".0" at the end: 

```rust
fn main() {
  let config = ser::PrettyConfig::new();
  let result = ser::to_string_pretty(&0.00000000000000005, config).unwrap();
  assert_eq!(result, "0.00000000000000005");
}
```

```
  left: `"0.00000000000000005.0"`,
 right: `"0.00000000000000005"`', src/main.rs:6:3
```

KirbyER:
Since `EPSILON` is defined as the difference between 1 and the next highest value, and values close to zero have higher precision, very small values were serialized with an extra ".0" e.g. "0.00000000000000005.0". These values then can't be properly deserialized. This fix compares the fractional part of the float to 0.0 instead of using `EPSILON`.

Fixes #374.

- [ ] I've included my change in `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix issue [#203](https://github.com/ron-rs/ron/issues/203) with full de error positioning ([#356](https://github.com/ron-rs/ron/pull/356))
 - Fix issue [#367](https://github.com/ron-rs/ron/issues/367) with eager implicit some ([#368](https://github.com/ron-rs/ron/pull/368))
 - Fix issue [#370](https://github.com/ron-rs/ron/issues/370) with `FromStr`-equivalent float EBNF and `Error::FloatUnderscore` ([#371](https://github.com/ron-rs/ron/pull/371))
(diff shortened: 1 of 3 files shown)


MomoLangenstein:
I found why the tests are failing in an unrelated way - for some reason `\0` now gets printed as `\0` instead of `\u{0}` (which was the case before), which is not yet supported. I'll add a quick fix so we can get this PR merged as well

MomoLangenstein (commented):
Could you file a short issue for this bug and then link that in a CHANGELOG entry? Otherwise, this looks good to me - thanks for catching the issue!

MomoLangenstein:
Thanks for opening the issue @KirbyER! Could you rebase your branch and add the following line (or a better description) to the bottom of the unreleased section of the CHANGELOG:
```
- Fix issue [#374](https://github.com/ron-rs/ron/issues/374) extraneous .0 for small floats ([#372](https://github.com/ron-rs/ron/pull/372))
```
Thanks again for catching this bug and finding a fix!

kvark (approved):
This makes sense!

KirbyER:
Sorry, got pulled away yesterday between opening the issue and updating the changelog. Done now. Thanks for sorting out the testing failure so quickly!

MomoLangenstein (approved):
LGTM

PR #373, opened 2022-03-31
Fix \0 in string on nightly + enable weekly CI

MomoLangenstein:
Add the `\0` special case to escape character handling to fix parsing it on nightly tests.

Also enables CI tests on every Sunday at midnight to catch nightly failures in quiet times.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,8 @@ on:
     branches:
     - master
     - 'v*.*'
+  schedule:
+    - cron: '0 0 * * 0'
 
 jobs:
   tests:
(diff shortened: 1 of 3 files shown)


MomoLangenstein:
?r @torkleyy This should be a very simple fix - I will merge the PR tonight to unblock #372 unless there is a need for a change

.github/workflows/ci.yaml
```diff
@@ -8,6 +8,8 @@ on:
     branches:
     - master
     - 'v*.*'
+  schedule:
+    - cron: '0 0 * * 0'
```
kvark, line 12: wow such fancy!

PR #378, opened 2022-04-25
added missing whitespace to named_field

raffimolero:
not 100% sure but i think named_field should allow whitespace after the colon and before the value in the EBNF

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -126,7 +126,7 @@ tuple = "(", [value, { comma, value }, [comma]], ")";
 struct = unit_struct | tuple_struct | named_struct;
 unit_struct = ident | "()";
 tuple_struct = [ident], ws, tuple;
-named_struct = [ident], ws, "(", [named_field, { comma, named_field }, [comma]], ")";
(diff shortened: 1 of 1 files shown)


MomoLangenstein:
That is correct - and it seems we might be missing even more `ws` there.

docs/grammar.md
```diff
@@ -127,7 +127,7 @@ struct = unit_struct | tuple_struct | named_struct;
 unit_struct = ident | "()";
 tuple_struct = [ident], ws, tuple;
 named_struct = [ident], ws, "(", [named_field, { comma, named_field }, [comma]], ")";
```
MomoLangenstein, line 129:
```suggestion
named_struct = [ident], ws, "(", ws, [named_field, ws, { comma, ws, named_field, ws }, [comma, ws]], ")";
```

raffimolero: the extra `ws` by the commas are unnecessary, `comma = ws, ",", ws`

MomoLangenstein: Ah sorry, you’re absolutely right. I think my edit should have corrected it now?

raffimolero: Looks good. Maybe comb the rest of the definitions as well.

Issue #365, opened 2022-02-01
Release 0.7.1 patch


kvark:
We need #335 to land on crates. There may potentially be other things that could be included.
@MomoLangenstein or @torkleyy - would you be interested in driving the patch release?

MomoLangenstein:
@kvark I'd be very happy to work on the release. As I haven't created and uploaded a release before, I might need some brief guidance.

I think an updated version of #334 should also be included (otherwise we'd move the `struct_names` option to `PrettyConfig` just to move it again in the next update). Since there hasn't been any progress on that PR, I could take care of rebasing its changes on the current main branch and moving the option to `Options`.

#363 would be a very small change that could be included / closed as well.

MomoLangenstein:
Edit: I had another think about #334 and at least for me I'm still not sure what the correct way there is. To get 0.7.1 to land quickly, it might be easier to just skip it and potentially move the `struct_names` config later.

I think it is correct that `struct_names` does more than just make the RON prettier (but so does `decimal_floats`, just to a lesser extent). Moving it to `Options`, however, also doesn't feel quite right as it only affects the `Serializer` so far. We could introduce a similar option for the `Deserializer` to enforce including struct names if enabled. Another option would be to go with the `_advanced` function pattern, but add new extensible config structs (like `PrettyConfig`) for both the `Serializer` and `Deserializer`, to allow them to more easily evolve should more options be needed in the future.

kvark:
Thank you for helping!

We'd need to summon @torkleyy about process here, since they [convinced me to change](https://matrix.to/#/!jlooIrYHEoEvaAyVet:matrix.org/$xxmqztvgd-6t9RNuHacCiAL35ZzjT3Kbn9gNNLju-eI?via=matrix.org) the release/contribution workflow some time ago relative to what I do for all other projects, which resulted in #327 .

In other projects we'd just back-port some of the commits onto v0.7 branch and publish a patch from there.

MomoLangenstein:
Yes, that would work as well

kvark:
Yeah, we'll be doing a version churn, causing downstream to diverge in the used RON versions and produce duplicates in the vendored trees. This is quite unfortunate. Are you sure the simplified workflow is worth it?

torkleyy:
Well, I don't think many libraries should be using RON, so I don't expect RON to be deep in a dependency tree.

kvark:
There is quite a few libraries - https://crates.io/crates/ron/reverse_dependencies
[...]

PR #360, opened 2022-01-26
Implement support for `DeserializeSeed` through `Options`

MomoLangenstein:
Fixes #359

* [ ] I've included my change in `CHANGELOG.md`

PR #382, opened 2022-06-06
Release v0.7.1

torkleyy:
Closes https://github.com/ron-rs/ron/issues/365

## TODO

- [x] Figure out what to do with `struct_names`
- [x] Changelog

---

@kvark @MomoLangenstein Only four months and four days later and I found some time to do the long awaited 0.7.1 release :sweat_smile: 

I've included all merged PRs since the v0.7.0 release, **except**:

- https://github.com/ron-rs/ron/pull/329
- https://github.com/ron-rs/ron/pull/351
- https://github.com/ron-rs/ron/pull/356
- https://github.com/ron-rs/ron/pull/361

## Changelog

- Add `struct_names` option to `PrettyConfig` ([#329](https://github.com/ron-rs/ron/pull/329))
- Fix newtype variant unwrapping around enum, seq and map ([#331](https://github.com/ron-rs/ron/pull/331))
- Implement `unwrap_newtypes` extension during serialization ([#333](https://github.com/ron-rs/ron/pull/333))
- Implement `unwrap_variant_newtypes` extension during serialization ([#336](https://github.com/ron-rs/ron/pull/336))
- Fix issue [#338](https://github.com/ron-rs/ron/issues/338) value map roundtrip ([#341](https://github.com/ron-rs/ron/pull/341))
- Fix issue [#289](https://github.com/ron-rs/ron/issues/289) enumerate_arrays comments ([#344](https://github.com/ron-rs/ron/pull/344))
- Report struct name in expected struct error ([#342](https://github.com/ron-rs/ron/pull/342))
- Add `Options` builder to configure the RON serde roundtrip ([#343](https://github.com/ron-rs/ron/pull/343))
- Fix issue [#359](https://github.com/ron-rs/ron/issues/359) with `DeserializeSeed` support ([#360](https://github.com/ron-rs/ron/pull/360))
- Fix issue [#370](https://github.com/ron-rs/ron/issues/370) with `FromStr`-equivalent float EBNF and `Error::FloatUnderscore` ([#371](https://github.com/ron-rs/ron/pull/371))
- Fix issue [#374](https://github.com/ron-rs/ron/issues/374) extraneous .0 for small floats ([#372](https://github.com/ron-rs/ron/pull/372))

## `struct_names` issue

Pull Request https://github.com/ron-rs/ron/pull/329 moved the `struct_names` from the `Serializer` to the `PrettyConfig`. This included a breaking change to `Serializer::new`, removing the third parameter. I see a couple ways to handle this:

### a) for 0.7.x, include the field in both `Serializer` and `PrettyConfig` and generate struct names if either are `true`

Advantages:

- easy to implement

Disadvantages:

- possibly surprising if `PrettyConfig` and the `Serializer::new` argument are controlled by different people and one sets it to `false`

### b) use the third parameter to override the field in `PrettyConfig`

Advantages:

- easy to implement
- won't change the behavior of any existing applications

Disadvantages:

- might be surprising for end-users only controlling the `PrettyConfig` if the `struct_names` property gets overwritten -> could besolved by doing `pretty_config.struct_names = pretty_config.struct_names || struct_names` (but that means `struct_names = false` will be ignored)
- non-pretty serialization included this option until now, but this makes `struct_names` no longer an option with compact serialization
- or, alternatively, we could try constructing a dummy `PrettyConfig` which ends up emulating non-pretty serialization

--> Were we aware that moving the field will no longer allow struct names with non-pretty serialization when merging this change?

### c) consider reverting the change

There are a couple things I don't like about this change:

- struct names can no longer be emitted with compact serialization
- pretty options can break functionality in unexpected ways if the RON is parsed by a different party using enums instead of structs for their `serde` structures or a different RON parser (seems fairly unlikely)

But I don't feel very strong about it. Having the field in `PrettyConfig` provides more comfort in most situations IMO, but the edge cases above could be a problem sometimes.


src/ser/mod.rs
```diff
@@ -231,20 +273,53 @@ impl Default for PrettyConfig {
...
+    /// # Deprecation
+    ///
+    /// This constructor takes `struct_names`, which has been moved to `PrettyConfig`.
+    /// Because of that, this constructor will overwrite the setting in
+    /// `PrettyConfig` to whatever you specify as a `struct_names` argument here.
+    #[deprecated(
+        note = "Serializer::new is deprecated because struct_names was moved to PrettyConfig"
+    )]
+    pub fn new(writer: W, config: Option<PrettyConfig>, struct_names: bool) -> Result<Self> {
+        todo!("figure out how to handle struct_names");
```
torkleyy, line 294: This needs to be resolved before merging.

MomoLangenstein:
Why were the other three PRs excluded (`struct_names` names sense) - are they breaking changes for v0.8?

torkleyy:
Yes exactly, they are breaking chanfges.

MomoLangenstein:
For v0.8 I think the `struct_names` would be a better fit for the `Options` since it might have an impact on parsing. This would likely be done by passing it into the constructor like it was before the PR, or by adding a separate post-construction modification method. So perhaps we could add it to `Options` already, add the post-construction method, and add a depreciation notice to the `Serializer` constructor that in the next version one should use `Options` or the post-construction modifier?

torkleyy:
> since it might have an impact on parsing

Hmm, it wouldn't right now, would it? I don't think there's a way to get the name of a struct when using `deserialize_any`, but this is just off the top of my head.
Despite that, I'm thinking it might still be the most consistent option.

kvark (commented):
Anything to resolve until CI passes and this is published?

torkleyy:
@kvark CI is failing intentionally, we only need to resolve how to handle `struct_names` as described in the PR description. Any thoughts on it?

torkleyy:
Okay, I fixed CI and implemented option a.

MomoLangenstein:
Ok, that looks good to me 👍 Perhaps we should document the planned depreciation in the release notes?

MomoLangenstein (approved):

MomoLangenstein:
LGTM

torkleyy:
Thanks for reviewing @MomoLangenstein

PR #385, opened 2022-06-19
Upper-bound dependencies to compile with MSRV 1.36

MomoLangenstein:
This is an example of how we could set upper version bounds for the published 0.7 version to work with its MSRV 1.36 without requiring fixes by the user.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,18 +27,6 @@ jobs:
           toolchain: ${{ matrix.rust }}
           profile: minimal
           override: true
-      - uses: actions-rs/cargo@v1
-        name: Downgrade bitflags to MSRV
-        if: ${{ matrix.rust }} == "1.36.0"
-        with:
-          command: update
-          args: -p bitflags --precise 1.2.1
-      - uses: actions-rs/cargo@v1
-        name: Downgrade indexmap to MSRV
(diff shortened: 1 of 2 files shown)


MomoLangenstein:
@torkleyy I have no idea if we could add this onto 0.7, e.g. as 0.7.2. Perhaps with some extra CI we could ensure that the current release version (or even the last two since the 0.6 branch also still exists) are also still tested (but probably allowing future clippy failures) is also tested as well and works with our official MSRV without needing user-fixes.

torkleyy:
I don't think we should restrict the dependency range like this, it will hurt users of newer Rust versions who will now end up with multiple versions of the same dependency (or unable to use the latest features of indexmap).

MomoLangenstein:
In that case, how do we ensure that both the MSRV and the latest stable work nicely? Perhaps we could document how you can fix such incompatibility errors, and give the explicit steps for the MSRV?

torkleyy:
Correct me if I'm mistaken, but can't you use the same procedure as we do in our CI step to ensure we locked the correct versions?
Binary crates using ron shouldn't have to worry about this as they'll have the old version locked anyways, and libraries depending on ron will have to patch their lock file in CI (in case they test on old Rust versions) and give similar instructions to their users.

MomoLangenstein:
Superseded by #386

PR #388, opened 2022-06-26
Add ci caching

ludicast:
This adds "official" [cargo caching](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) to cut CI time (from ~3 mins to ~1 mins). We also bump the version of github's checkout action to the latest.

Screenshot from 2022-06-26 06-50-17
to
Screenshot from 2022-06-26 07-20-56

Note that the codecov run failures are just because I don't have access to the api of course. If preferred, I can extract/isolate the cache & toolchain steps into a shared local action that this imports, [similar to askama's](https://github.com/djc/askama/blob/main/.github/actions/setup/action.yml).

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,19 @@
+name: Setup Rust Environment
+
+inputs:
+  key:
+    description: Cache key
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+        key: ${{ inputs.key }}-cargo-${{ hashFiles('Cargo.lock') }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,7 +90,7 @@ jobs:
           profile: minimal
           components: llvm-tools-preview
           override: true
-      - run: cargo install grcov --locked
+      - run: cargo install grcov --force --locked
       - run: |
           cargo test --all-features --all-targets
           grcov . -s . --binary-path ./target/debug/ \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,16 @@ jobs:
         rust: [1.46.0, stable, nightly]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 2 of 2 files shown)


MomoLangenstein:
Thanks for the PR - I just our CI twice to see the impact:

- Clippy: 1m 34s -> 25s
- Format: 13s -> 9s
- Coverage: 5m 1s -> 3m 12s
- Tests nightly: 3m 42s -> 1m 23s
- Tests stable: 4m 19s -> 1m 23s
- Tests 1.46: 4m 19s -> 1m 19s

MomoLangenstein:
Perhaps we could get the codecov time down further by removing `--locked` from the action? I’d also be open to using a shared local option to combine the caching steps.

ludicast:
> Perhaps we could get the codecov time down further by removing `--locked` from the action? I’d also be open to using a shared local option to combine the caching steps.

cool @MomoLangenstein  - combined caching in a single shared action and removed `--locked` -.  hopefully it speeds up the coverage job

MomoLangenstein:
> > Perhaps we could get the codecov time down further by removing `--locked` from the action? I’d also be open to using a shared local option to combine the caching steps.
> 
> cool @MomoLangenstein - combined caching in a single shared action and removed `--locked` -. hopefully it speeds up the coverage job

Ah, I made a mistake earlier - we might need to keep the `--locked` but should remove the `--force` … sorry about that :)

ludicast:
> > > Perhaps we could get the codecov time down further by removing `--locked` from the action? I’d also be open to using a shared local option to combine the caching steps.
> > 
> > 
> > cool @MomoLangenstein - combined caching in a single shared action and removed `--locked` -. hopefully it speeds up the coverage job
> 
> Ah, I made a mistake earlier - we might need to keep the `--locked` but should remove the `--force` … sorry about that :)

np - switched them out

MomoLangenstein:
Huh, this seems more difficult than I had thought. If you know a way to use `--locked` during the install AND cache the install then we should do that. Otherwise, we could merge the working version with both `--locked` and `--force` for now and postpone a solution for the installation.

ludicast:
> Huh, this seems more difficult than I had thought. If you know a way to use `--locked` during the install AND cache the install then we should do that. Otherwise, we could merge the working version with both `--locked` and `--force` for now and postpone a solution for the installation.

Yeah, there's probably a clear way to do it, but not sure r/n. Def can do as a fast-follow, so undid for now.

MomoLangenstein:
Thanks for your work @ludicast! With these initial improvements in place, we can now see if we can get the coverage times down as well in a follow-up.

PR #390, opened 2022-06-27
add target dir for code coverage job

ludicast:
Using alternative install directory that should be caught up in the caching (`CACHE_TARGET_DIR` envvar mentioned [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html)). 

This also removes `nightly` from the cache key to blow out the cache (since I believe this change doesn't affect the cargo.lock sha). IMHO the disambiguation isn't necessary (don't imagine coverage will ever run from a matrix of rust versions).


MomoLangenstein (approved):
LGTM

Issue #393, opened 2022-07-24
Enhancement: better error messages for missing struct fields?


d86leader:
Currently missing fields in struct gives no line info, not even the typename info, which makes it hard to find where in the row of same-y declarations you missed a field. For example:

```rust
#[derive(serde::Deserialize, Debug)]
struct Test {
    f1: i64,
    f2: String,
}

fn main() {
    let s = r#"
Test (
    f1: 55,
)
"#;
    let x: Result<Test, _> = ron::from_str(s);
    println!("{:?}", x);
}
```
gives: `Err(Error { code: Message("missing field `f2`"), position: Position { line: 0, col: 0 } })`

It would be great if I could get line info, or at least the name "Test" of the record.

Also there is a similar problem with enum variants. Which seems doubly strange, as when I write an incorrect constructor for a struct, like replacing "Test" with "NotTest" in the example above, I get precise error location and type info.

MomoLangenstein:
On `master`, #356 is already included (this was excluded from `v0.7.1` in #382), which does give the correct line info (also for enum variants):
```rust
Err(SpannedError { code: Message("missing field `f2`"), position: Position { line: 4, col: 1 } })
```
I'll check if we could also give the name of the struct or enum variant where the field is missing

torkleyy:
I think we should make necessary API breaking changes now (if any) and then release master as 0.8, so the new features, especially better errors, become available.

MomoLangenstein:
Ok, so the error message here comes from `serde`'s default implementation of [`Error::missing_field`](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.missing_field), which just generates this string. If we wanted, `ron` could overwrite all or some of the `Error` methods to capture the results in nicer error variants - this might be a good idea on its own, just to make error handling easier.

If we then have such variants, we could populate them with extra optional information that would be `None` by default, e.g. the name of the surrounding struct, the name of the field etc. While this might make for fantastic error messages in some cases, it could also bloat the error type. For now, I don't think adding this functionality to `ron` is necessary.

What I use for such nice contextual error messages is the [`serde_path_to_error`](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/) crate:
```rust
let mut de_ron = ron::Deserializer::from_str_with_options(ron_str, ron_options)?;

let mut track = serde_path_to_error::Track::new();
let de = serde_path_to_error::Deserializer::new(&mut de_ron, &mut track);

let result = match deserializer(de) {
[...]

PR #394, opened 2022-07-24
Improve error information for serde errors

MomoLangenstein:
Followup to #393

Adds explicit variants in `ron::Error` for `serde` errors from [`serde::de::Error`](https://docs.rs/serde/latest/serde/de/trait.Error.html) (`serde::ser::Error` only has the `Error::custom` method).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,7 +3,7 @@ use std::{error::Error as StdError, fmt, io, str::Utf8Error, string::FromUtf8Err
 
 /// This type represents all possible errors that can occur when
 /// serializing or deserializing RON data.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct SpannedError {
     pub code: Error,
     pub position: Position,
@@ -12,7 +12,7 @@ pub struct SpannedError {
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 pub type SpannedResult<T> = std::result::Result<T, SpannedError>;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Error {
     Io(String),
@@ -59,13 +59,9 @@ pub enum Error {
     Utf8Error(Utf8Error),
     TrailingCharacters,
 
-    ExpectedDifferentType {
-        expected: String,
-        found: UnexpectedSerdeTypeValue,
-    },
     InvalidValueForType {
         expected: String,
-        found: UnexpectedSerdeTypeValue,
+        found: String,
     },
     ExpectedDifferentLength {
         expected: String,
@@ -148,16 +144,6 @@ impl fmt::Display for Error {
             }
             Error::UnexpectedByte(ref byte) => write!(f, "Unexpected byte {:?}", byte),
             Error::TrailingCharacters => f.write_str("Non-whitespace trailing characters"),
-            Error::ExpectedDifferentType {
-                ref expected,
-                ref found,
-            } => {
-                write!(
-                    f,
-                    "Expected a value of type {} but found {} instead",
-                    expected, found
-                )
-            }
             Error::InvalidValueForType {
                 ref expected,
                 ref found,
@@ -244,7 +230,7 @@ impl fmt::Display for Error {
     }
 }
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Position {
     pub line: usize,
     pub col: usize,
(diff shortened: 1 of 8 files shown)


MomoLangenstein:
?r @torkleyy

MomoLangenstein:
This is the first proof-of-concept. Notably, I've also ported `serde`'s [`Unexpected`](https://docs.rs/serde/latest/serde/de/enum.Unexpected.html) enum, which I'm not quite sure about. It does give us slightly nicer error messages and would allow users to match on the unexpected value (though I'm not quite sure if anyone would need that). However, it also adds some very `serde`-specific terms to our codebase, which don't map quite so well to `ron`. We could decide to just remove that part and stringify the unexpected values as `serde` does, or we could limit the enum to entries that make sense for us (we'd still adopt `serde`'s decision to only store `[u|i]64`).

MomoLangenstein:
What I haven't touched yet is the area of adding more contextual information to those new errors. Something we might be able to do is to decorate the `deserialize_enum` and `deserialize_struct`-like methods to check for the new errors and insert struct/enum name information.

We could probably also try to find the name of the current field, though in my opinion, that would be overkill and is already handled by crates like [`serde_path_to_error`](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/).

MomoLangenstein:
I've tried to do the struct/enum/variant name extraction for even better error messages. Just extracting struct/enum names is easy, and we can store them as `&'static str`. For variant names, however, which are important for struct variants which to the `ron`-writing user look just like structs, I had to add a small hack. The `Deserializer` now remembers what the last deserialized identifier was. Since enums in `ron` use identifiers for the variants, I can check for that last one and assume it was the variant id. This does provide better errors (for externally tagged enums, internally tagged ones cannot get any context since they are based on `deserialize_any`, adjacently tagged and untagged enums get ok-ish errors), but means that the struct/enum/variant name is stored as a `String`.

MomoLangenstein:
@torkleyy @d86leader What are your thoughts?

src/error.rs
```diff
@@ -58,6 +58,37 @@ pub enum Error {
...
     TrailingCharacters,
+
+    ExpectedDifferentType {
+        expected: String,
+        found: UnexpectedSerdeTypeValue,
+    },
+    InvalidValueForType {
+        expected: String,
+        found: UnexpectedSerdeTypeValue,
+    },
```
d86leader, line 69: What's the difference between those two errors? In fact, I can't understand at all what the "DifferentType" means. (:
When deserializer expects an integer, but I provide a string, which error is raised?

MomoLangenstein: The only difference is that they come from different error methods in `serde`: [`invalid_type`](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.invalid_type) and [`invalid_value`](https://docs.rs/serde/latest/serde/de/trait.Error.html#method.invalid_value). When your field is an integer but you give a string, neither will be called since `ron` handles that one natively. It is actually quite difficult to observe that one in the wild. The `invalid_value` error is more common, e.g. when you try to deserialise `0` into `NonZero`.

MomoLangenstein: Perhaps it would be useful to go over the existing errors and reclassify them as type or value errors as well so there's no overlap - this could also allow us to clean up those two errors from `serde` based on what we think they should be

torkleyy (approved):
Looks good to me! I'm not entirely sure if we should expose provide api guarantees for all of our erroor types. serde_json doesn't expose much in that regard.

torkleyy:
I will have to give your comments a closer look later this week, I'm busy right now, sorry. But feel free to proceed without me if you.
I think workarounds are fine if they don't overly complicate the codebase and provide useful information, but I'd be cautious of exposing those details via an API because there's a higher chance we'll change them in the future. In general I think we shouldn't expose the full error enum as we do know.

MomoLangenstein:
@torkleyy I've opted to simplify the error enum a bit by merging `serde`'s `invalid_type` and `invalid_value` into the same `InvalidValueForType` type in `ron`. Furthermore, this error now contains the stringified version of `found` field instead of replicating `serde`'s `de::Unexpected` type (however, I still implement some custom formatting for the conversion to get slightly nicer error messages). This now seems like a good compromise to me where we get some extra info from `serde`'s useful errors like the missing field or unknown variant ones without having to also include its expected vs unexpected system.

MomoLangenstein:
CI is blocked on https://github.com/serde-rs/serde/issues/2255

torkleyy:
Since we're doing a API-breaking release, I think we should use the opportunity to bump MSRV and get rid of the CI problem at the same time :)

MomoLangenstein:
> Since we're doing a API-breaking release, I think we should use the opportunity to bump MSRV and get rid of the CI problem at the same time :)

@torkleyy I've opened #396 to bump the MSRV to 1.56. Once that PR is merged I'll rebase this one off it.

MomoLangenstein:
@torkleyy This PR is now ready to be merged if you're happy with the changes :)

torkleyy (approved):
Looks great, can't wait to use ron with such nice errors :)

PR #400, opened 2022-08-16
Expected bytes error and import cleanup

juntyr:
This cleanup PR includes
- a small change to the error message for the unexpected bytes error to print the bytesrting with base64, since that's how it shows up in the ron
- reformatting of all import statements (using the unstable `imports_granularity = "Crate"` and `group_imports = "StdExternalCrate"` options)


MomoLangenstein:
The codecov changes seem to be spurious since they only affect `use` statements.

PR #402, opened 2022-08-20
Fixed raw struct name identifier parsing

### LINKED ISSUE #401: Raw struct names cannot be deserialised
```rust
use ron::Error;

#[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
#[serde(rename = "Hello+World")]
struct RawStruct {
    #[serde(rename = "ab.cd-ef")]
    field: bool,
}

#[test]
fn test_raw_identifier_roundtrip() {
    let val = RawStruct { field: true };

    let ser =
        ron::ser::to_string_pretty(&val, ron::ser::PrettyConfig::default().struct_names(true))
            .unwrap();
    assert_eq!(ser, "r#Hello+World(\n    r#ab.cd-ef: true,\n)");

    // Fails with an error that "Hello+World" was expected but "Hello+World" was found
    let de: RawStruct = ron::from_str(&ser).unwrap();
    assert_eq!(de, val);
}

```

juntyr:
Fixes #401

- adds documentation about valid identifiers to the grammar
- checks identifier validity upon serialisation
- fix struct name parsing to support raw identifiers

@torkleyy Deserialisation currently does NOT validate expected identifiers, so one could get an error saying `found "World" but expected "Hello World"` which has absolutely no user-side fix if you don't control the data structure you're deserialising into. I could specialise the error messages to add a note if an invalid identifier was expected, but that might clutter the code base.

### DIFF
--- a/tests/401_raw_identifier.rs
+++ b/tests/401_raw_identifier.rs
@@ -11,6 +11,7 @@ struct EmptyStruct;
 
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
 #[serde(rename = "Hello+World")]
+#[serde(deny_unknown_fields)]
 struct RawStruct {
     #[serde(rename = "ab.cd-ef")]
     field: bool,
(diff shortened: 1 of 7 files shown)


tests/401_raw_identifier.rs
```diff
@@ -0,0 +1,83 @@
...
+        SpannedError {
+            code: Error::ExpectedDifferentStructName {
+                expected: "Hello World",
+                found: String::from("Hello"),
+            },
+            position: Position { line: 1, col: 6 },
+        }
+    );
+
+    let de = ron::from_str::<EmptyStruct>("").unwrap_err();
```
torkleyy, line 54: Can we add a test for an empty raw identifier? Would that get accepted?

MomoLangenstein: Done ... though this exploded a bit in d85ca35a13bf2ebccd16237479adfafc8293182f and 37df3d8f6a17b3adf0ee0b2fbf668d972710c80b

Dealing with empty (raw) identifiers is tricky. The solution now is the following:
- if the expected identifier is valid, we complain about not finding the opening `(` for the expected named struct
- if the expected identifier is empty, the same error struct is used but the message is adapeted to the following: `1:1: Expected only opening `(`, no name, for un-nameable struct`
- if the expected identifier is invalid in another way, the usual error message (as in case 1) is given, but the expected struct name is printed as `"ö"_[invalid identifier]`

I've added this new identifier formatting to error messages in general, such that struct, variant, and field names are now always automatically printed either as \`id`, \`r#i+d`, or "ö"_[invalid identifier]

MomoLangenstein: This now also does a tiny bit of cleanup to the `deserialize_struct` method by pulling out everything after the struct name check, so that we no longer pass `""` as an identifier to the method internally, meaning it can now complain about it as an error instead of having to assume it was our internal code that passed it in

MomoLangenstein:
@torkleyy Can I merge this PR?

torkleyy (approved):
Yes, looks good to me :+1:

PR #403, opened 2022-08-21
Fixes #115 struct flattening

### LINKED ISSUE #115: Struct flattening not working

juntyr:
Fixes #115 by allowing `ron` maps to be deserialised from structs. This is a hack. I am very unsure if this is something we want `ron` to support with `serde`'s current model, or if we wanted to hold out for some extension to `serde` that would give `Deserializer`'s more control about how to handle things like flatten and untagged enums etc.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -495,6 +495,43 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
+        // Enable serde flatten by checking for a struct on a working copy
+        let mut wc = self.bytes;
+
+        // Remove any leading struct name identifier
+        if !self.newtype_variant {
+            let _ = wc.identifier();
+            wc.skip_ws()?;
+        }
+
+        // Check if we see something struct- or tuple-like
+        if self.newtype_variant || wc.consume("(") {
+            wc.skip_ws()?;
+
+            let has_field_ident = wc.identifier().is_ok();
+
+            wc.skip_ws()?;
+
+            // It's a struct if it has either a field name followed by a colon
+            //  or the struct is immediately closed
+            let is_struct = wc.eat_byte().map_or(false, |c| {
+                if has_field_ident {
+                    c == b':'
+                } else {
+                    c == b')'
+                }
+            });
+
+            if is_struct {
+                // Remove any leading struct name identifier
+                if !self.newtype_variant {
+                    let _ = self.bytes.identifier();
+                }
+
+                return self.deserialize_struct("", &[], visitor);
+            }
+        }
+
         self.newtype_variant = false;
 
         if self.bytes.consume("{") {
--- a/tests/115_struct_flattening.rs
+++ b/tests/115_struct_flattening.rs
@@ -1,0 +1,170 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, PartialEq, Eq, Debug)]
+struct Main {
+    #[serde(flatten)]
+    required: Required,
+    #[serde(flatten)]
+    optional: Optional,
+
+    some_other_field: u32,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Eq, Debug)]
+struct Required {
+    first: u32,
+    second: u32,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Eq, Debug)]
+struct Optional {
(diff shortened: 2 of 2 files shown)


MomoLangenstein:
@torkleyy thoughts?

torkleyy:
Yeah I'm also unsure what drawbacks this hack might have. I think there always will be some limitations when using RON + serde (which is the only option today), simply because serde was designed for JSON. IIRC there's also untagged enums that won't work with ron.

The question is is it worth having this feature? Could it prevent using someone from ron because a dependency crate uses struct flattening? I think probably not. The most common use of flatten I see is for all the ignored fields to store them in a map. I think we do support that, right?

MomoLangenstein:
> Yeah I'm also unsure what drawbacks this hack might have. I think there always will be some limitations when using RON + serde (which is the only option today), simply because serde was designed for JSON. IIRC there's also untagged enums that won't work with ron.

I once did a similar experiment for untagged enums, and they could sort of be supported by ron if `ron::Value` subscribes to `serde`‘s ideas (structs and enums become single-item maps). Both of these solutions are in no way pretty. Imo `serde` should allow `Deserializer` to supply its own `Value` type that can then get special support, and have special methods to handle all the enum encodings etc within the data format itself. In the absence of such support, we’ll have to decide and document which feature of serde we do guarantee support for.

> The question is is it worth having this feature? Could it prevent using someone from ron because a dependency crate uses struct flattening? I think probably not. The most common use of flatten I see is for all the ignored fields to store them in a map. I think we do support that, right?

I thought so too, but unfortunately both use cases call `deserialize_map`, so it seems that we either support no flatten or all flatten.

torkleyy:
I see. Well I don't think `serde` will implement any of that, looking at the issues it's doing bug fixes only.

MomoLangenstein:
Re untagged enums: https://github.com/ron-rs/ron/compare/master...JuniperLangenstein:253-untagged-enums

MomoLangenstein:
I also just thought of something else: in that prototype branch I made a while ago, struct-like things were deserialised as single-item maps into `Value`. This allows us to do the following round trip: Rust (-> ron) -> Value (-> ron -> Value) -> Rust. However, we cannot do Rust -> Value -> ron -> Rust, since the Value does a lot of heavy lifting to interpret those single maps.

We could, in theory, make single-item maps special in the Deserialize impl for Value and convert them into a ron-specific struct-like type immediately. This would allow us to do Rust -> Value -> ron -> Rust, and the Value would be much more like a ron AST than a JSON AST.

However, this would mysteriously break any single-item maps in user code that go through Value. We could, again, special case that (though this would be a bit ugly and would add even more edge cases), but as soon as such a Value would be written to ron, we'd be done. Unless we special-case the ron Deserializer impl as well. However, at this point we've got a paradox: flatten needs to interpret `Struct(a: 2)` as `{"a": 2}`, but our value case needs to interpret it as `{"Struct": {"a": 2}}`.

Hence, I think we have two options going forward with Values:

1. Stick with a JSON-AST-like Value but ensure it supports all enum shenanigans that serde offers, it can round trip through other formats, but Rust->Value->ron->Rust doesn't work

2. Have a ron-AST-like Value that supports all enum shenanigans that serde offers, it cannot round trip through other formats, and single-item maps don't round trip through Value

Even though a ron-AST-like Value would be nicer, I think if we want to have the best compatibility, we'd have to stick with the JSON-AST-like Value, but we could offer some convenience inspector methods to check if the Value looks e.g. like a self-describing enum

juntyr:
Superseded by #455

PR #407, opened 2022-09-11
Implemented ron::value::RawValue

MomoLangenstein:
Adds the `ron::value::RawValue` type, inspired by [`serde-rs`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html), which holds a valid RON fragment string and can be (de)serialised inside any container.


MomoLangenstein:
?r @torkleyy I stumbled across this API some time ago and thought it would be quite useful to also have it inside `ron`. While I added some convenience functions to `RawValue` for the common cases, I decided against implementing `Serializer` and `Deserializer` directly as those would only duplicate functionality which is already available by just using the existing `Deserializer` with the `&str` one can get from `RawValue`.

torkleyy (approved):
Makes sense & looks good to me :)

Issue #397, opened 2022-08-15
`Value` meta-issue


MomoLangenstein:
Just to start tracking all the `Value` related ones - I'll try to tackle them in 0.9:

- [ ] #357
- [ ] #255
- [ ] #217
- [ ] #189
- [ ] #140 
- [ ] #122 
- [ ] #115

MomoLangenstein:
I've been thinking about the `ron` `Value` for a few days and wanted to summarise my thoughts:

- the `Value` type should be extended to be more like a `ron` AST (extensions excluded?)
- most issues stem from `ron` having a type system (in the data format) that is both more and less expressive than `serde`'s `de::Visitor` API. On the one hand, `ron` has named units, (named) tuples, (named) struct-likes, all in addition to `serde`'s sequences and maps. On the other hand, it does not distinguish between structs and enums.
- serialising into a `Value` should be no problem, but it might be useful to not make `Value` a `Serializer` itself but instead reuse the existing `Serializer` to benefit from the extension handling (i.e. serialising into a `Value` should give the `ron` as serialising to a `String`)
- deserialising from an extended `Value` should present little problems, but would also benefit from plugging into the existing `Deserializer` to handle the unwrap extensions. That way, all non-self-describing methods should work flawlessly
- `deserialize_any` could be adapted to support `serde`'s untagged functionality ... to some extent (the unwrap extensions will never work with it since it requires support from our `Deserializer`) ... perhaps this should be a separate issue
- transforming a `Value` into a `ron` string is trivial, but could perhaps be integrated with the `Serializer` to support the same `PrettyConfig`
- implementing `Deserialize` for `Value` is the difficult part, since `serde` just doesn't give us sufficient information for it to work losslessly. Perhaps we should, therefore, split the functionality into (1) a `Value` can be serialised and deserialised by any data format - but this does not mean you can deserialise a `Value` from a `ron` string, and (2) a `Value` can be parsed from a `ron` string. This change would be **breaking** since `Value`'s `Serialize` and `Deserialize` impls would do something very different from before, while the functionality of going between `ron` and `Value` would be exposed in separate methods.

@torkleyy what are your thoughts on this?

PR #409, opened 2022-09-18
Explore untagged enums

juntyr:
Experimentation for #253 and #397 

### DIFF
--- a/src/de/id.rs
+++ b/src/de/id.rs
@@ -178,10 +178,19 @@ impl<'a, 'b: 'a, 'c> de::Deserializer<'b> for &'c mut IdDeserializer<'a, 'b> {
         unimplemented!("IdDeserializer may only be used for identifiers")
     }
 
-    fn deserialize_newtype_struct<V>(self, _: &'static str, _: V) -> Result<V::Value>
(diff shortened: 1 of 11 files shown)


MomoLangenstein:
@torkleyy This is mostly an old experiment that I just rebased, but I wanted to see what would need to be done to make this feel right. Essentially, we could land this change 'as-is' to support all `serde` enum shenanigans. While this might be an ok first step, it would change how our `Value` type works and tie it even more to `json`. Hence, I'm now experimenting a bit with whether this change could be done in combination with a more `ron`y `Value` that has variants for structs (and tuples). Here are my findings so far:

* since serialising requires `&'static str` for all struct and field names, the `Value` would need to use those as well. Hence, any deserialising would require some string interning (e.g. using https://github.com/CAD97/simple-interner which supports `&'static str` from rust 1.63 onwards)
* structs without a name in `ron` will bite us - we can either reject deserialising them or reject serialising them since `serde` doesn't have that concept
* so far treating structs like enums in `ron` seems to work fine

I think in the end this will come down to whether we want the `ron::Value` to be essentially a clone of `serde_json::Value` (easy to implement but no roundtrips) or reflect `ron` more (more work). Any thoughts?

juntyr:
Superseded by #451

PR #413, opened 2022-10-04
[v0.9] Breaking: Treat Some like any newtype variant

juntyr:
How should `Some` be treated in the presence of the `unwrap_variant_newtypes` extension? With that extension, `NewtypeVariant(Some(a: 42, b: 24))` already worked, but `Some(a: 42, b: 24)` did not. In my opinion `Some` is just a newtype variant and should parse consistently here. @torkleyy what are your thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `ron::value::RawValue` helper type which can (de)serialize any valid RON ([#407](https://github.com/ron-rs/ron/pull/407))
 - Fix issue [#410](https://github.com/ron-rs/ron/issues/410) trailing comma parsing in tuples and `Some` ([#412](https://github.com/ron-rs/ron/pull/412))
 - Error instead of panic when deserializing non-identifiers as field names ([#415](https://github.com/ron-rs/ron/pull/415))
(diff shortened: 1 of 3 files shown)


torkleyy (commented):
I think the change makes sense, but I'm wondering if it will break anything in practice. In any case this change needs to be in a breaking version release. Maybe we should merge it when we are about to release other breaking changes.

MomoLangenstein:
> I think the change makes sense, but I'm wondering if it will break anything in practice. In any case this change needs to be in a breaking version release. Maybe we should merge it when we are about to release other breaking changes.

That’s an interesting idea as well … what would the cost be of merging it right now but excluding it from a potential 0.8.1 release? I have to admit one motivation of this comes from the fuzzing results where this bug was rediscovered.

MomoLangenstein:
> I think the change makes sense, but I'm wondering if it will break anything in practice. In any case this change needs to be in a breaking version release. Maybe we should merge it when we are about to release other breaking changes.

Good point! I agree that we can keep around this change until we’re ready for 0.9 as it’s quite small and won’t be difficult to rebase then

juntyr:
#465 shows that this might actually be a bad idea, as `Some(a: 42)` cannot be handled (without very big and hacky stretches) inside `deserialize_any`.

juntyr:
#465 now includes and supersedes this PR

PR #416, opened 2022-10-07
Do not pin Cargo.lock for fuzzing

MomoLangenstein:
Cargo.lock pinned an outdated version of ron and its dependencies


MomoLangenstein:
@5225225 was there a particular reason for including the lock file?

torkleyy (approved):

5225225:
IIRC this is the generated gitignore that cargo-fuzz makes, unless I messed it up at some point in making it. Maybe the generated gitignore should ignore the lockfile? It *is* basically a binary, and in general the guidance is to commit lockfiles for binaries. Though that's for reproducable builds which matters less here.

in general I prefer to keep lockfiles so you can always check out even old commits and know they will build, but for a fuzzer I suppose it doesn't matter too much.

But it shouldn't be actually using an outdated version of ron because it should be a path dependency, so the version there is always up to date with what the code in-repo is.

I'm not too bothered either way, though.

MomoLangenstein:
> IIRC this is the generated gitignore that cargo-fuzz makes, unless I messed it up at some point in making it. Maybe the generated gitignore should ignore the lockfile? It _is_ basically a binary, and in general the guidance is to commit lockfiles for binaries. Though that's for reproducable builds which matters less here.
> 
> in general I prefer to keep lockfiles so you can always check out even old commits and know they will build, but for a fuzzer I suppose it doesn't matter too much.
> 
> But it shouldn't be actually using an outdated version of ron because it should be a path dependency, so the version there is always up to date with what the code in-repo is.
> 
> I'm not too bothered either way, though.

Thanks for sharing your insights @5225225! We could keep the lock file around but perhaps update it every so often? I only noticed this as I was running `cargo fuzz` and suddenly had a dirty file on my hands :)

PR #420, opened 2022-10-09
Add explicit recursion limits to avoid stack overflows

### LINKED ISSUE #307: Stack overflow when parsing deeply nested map
Reproduction:

```rust
fn main() {
    let _: Result<ron::Value, _> = ron::from_str(&"{".repeat(10_000));
}
```

Found by fuzzing, I'll make a distinct PR to add the fuzzer.

juntyr:
Fixes #307 and the long-standing potential for stack overflows while serialising or deserialising.

Now, `ron::Options` is by default configured for a recursion limit of 128 (here recursion is very loosely defined). This limit can be adjusted or disabled.

This bug was first reported by @5225225 in #307 and later by oss-fuzz.

This PR also slightly extends the `from_str` fuzz target to check serialising as well.


MomoLangenstein:
@torkleyy Does this API change look good to you? I would be surprised if anyone is using such highly recursive data structures in the wild and would be affected by the default limit. Hence, this might be a v0.9 change?

MomoLangenstein:
Part of the codecov regression is spurious, part is expected. As I've spent the past few commits increasing codecov this is unfortunate, but I'll get to adding coverage for the stack overflow checks at some point :)

MomoLangenstein:
@torkleyy As you might be busy, I will merge this PR on Friday if you don’t object. While this change would break deeply nested structures, I think this can be dealt with by pushing the change to 0.9 instead of 0.8.1. Apart from a new error, the API update is a non-breaking change.

torkleyy (approved):
Yes limiting the recursion limit is certainly useful especially for untrusted inputs. 👍

PR #431, opened 2022-12-12
Bump MSRV to 1.57.0 and base64 to 0.20

juntyr:
Supersedes #430

If we want to upgrade to `base64` 0.20, we need to bump our MSRV to 1.57 as well.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        rust: [1.56.0, stable, nightly]
+        rust: [1.57.0, stable, nightly]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


juntyr:
Will need to be followed up by a migration to 0.21 (#432)

torkleyy (approved):
Looks good, I only wonder if `base64` is good to have baked into RON, given that you could easily do it yourself with serde attributes.

Removing it would be a rather unpleasent breaking change for anyone using it however... Any thoughts?

You can merge it as-is, this is just what came to my mind right now ^^

juntyr:
> Looks good, I only wonder if `base64` is good to have baked into RON, given that you could easily do it yourself with serde attributes.

Do you mean users could hardcode how they want to (de)serialise their bytes? That's definitely true (and what I do in a personal use case), but we'd still need to figure out how we want to encode bytes in the ron format. Do you have a suggestion for what would be a good alternative?

> Removing it would be a rather unpleasent breaking change for anyone using it however... Any thoughts?

That's my concern too. Anyone using ron with byte-encoded data would suddenly no longer be able to parse their config files. Since that would be a breaking change in the format and not in the crate, I think it would be difficult to justify outside of a major (major) version bump.

> You can merge it as-is, this is just what came to my mind right now ^^

Thanks :)

PR #433, opened 2023-01-10
Bump base64 to v0.21

juntyr:
I've bumped base64 to 0.21 and refactored its usage by defining the used engine, `STANDARD`, as a private const in one place.

Supersedes #432


juntyr:
?r @torkleyy This is read to me merged if you give the ok

torkleyy (approved):

sdroege:
Not that it matters much, but you might want to look at the `data-encoding` crate as an alternative. It provides a simpler API.

juntyr:
> data-encoding

Thank you for the recommendation - `data-encoding`'s API definitely looks quite nice and we'll keep the crate in mind.

PR #434, opened 2023-01-16
Switch from `base64` to `data-encoding`

sdroege:
It offers the same functionality and provides a simpler API that provides exactly what is needed here.

### DIFF
--- a/src/ser/tests.rs
+++ b/src/ser/tests.rs
@@ -121,6 +121,16 @@ fn test_byte_stream() {
         "(0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15)"
     );
 
+    // Requires padding of the base64 data
+    let large = vec![0x01, 0x02, 0x03, 0x04];
+    let large = serde_bytes::Bytes::new(&large);
+    assert_eq!(to_string(&large).unwrap(), "\"AQIDBA==\"");
+
+    // Requires no padding of the base64 data
+    let large = vec![0x01, 0x02, 0x03, 0x04, 0x05, 0x06];
+    let large = serde_bytes::Bytes::new(&large);
+    assert_eq!(to_string(&large).unwrap(), "\"AQIDBAUG\"");
+
     let large = vec![255u8; 64];
     let large = serde_bytes::Bytes::new(&large);
     assert_eq!(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `escape_strings` option to `PrettyConfig` to allow serialising with or without escaping ([#426](https://github.com/ron-rs/ron/pull/426))
 - Bump MSRV to 1.57.0 and bump dependency: `base64` to 0.20 ([#431](https://github.com/ron-rs/ron/pull/431))
 - Bump dependency `base64` to 0.21 ([#433](https://github.com/ron-rs/ron/pull/433))
+- Switch from `base64` to `data-encoding` ([#434](https://github.com/ron-rs/ron/pull/434))
 
 ## [0.8.0] - 2022-08-17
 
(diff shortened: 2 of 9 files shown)


juntyr:
?r @torkleyy

sdroege:
> Does `data-encoding` provide the same base64 encoding scheme as the `base64` crate did?

Yes, and like `base64` it also provides a few others under different names.

> Could we perhaps add some tests to ensure that this would not break any parsing?

Sure, anything specific you're thinking of here or just a few encode/decode tests that contain base64 encoded bytes?

juntyr:
> > Does `data-encoding` provide the same base64 encoding scheme as the `base64` crate did?
> 
> Yes, and like `base64` it also provides a few others under different names.
> 
> > Could we perhaps add some tests to ensure that this would not break any parsing?
> 
> Sure, anything specific you're thinking of here or just a few encode/decode tests that contain base64 encoded bytes?

I think one roundtrip test for all three places where base64 is used should be good + one to ensure we are really using the standard base64 encoding with padding (i.e. the equivalent of https://docs.rs/base64/latest/base64/engine/general_purpose/constant.STANDARD.html)

sdroege:
> using the standard base64 encoding with padding

Yeah that's exactly it :)

I'll add some tests later!

src/ser/tests.rs
```diff
@@ -121,6 +121,16 @@ fn test_byte_stream() {
...
+    assert_eq!(to_string(&large).unwrap(), "\"AQIDBA==\"");
+
+    // Requires no padding of the base64 data
+    let large = vec![0x01, 0x02, 0x03, 0x04, 0x05, 0x06];
+    let large = serde_bytes::Bytes::new(&large);
+    assert_eq!(to_string(&large).unwrap(), "\"AQIDBAUG\"");
+
     let large = vec![255u8; 64];
     let large = serde_bytes::Bytes::new(&large);
     assert_eq!(
```
sdroege, line 136: The one just below is already would fail if the URL-safe version instead of the standard version is used.

Let me know if you have any other strings / kind of strings you'd like to have tested. Padding handling would be covered with the new tests now, and they pass before/after.

juntyr: Those look good - could you add one for the bytes error case as well where we use base64 to format the erroneously given bytes?

sdroege: I'm not entirely sure if you meant this, but what I just added seems like a good addition too.

juntyr: Absolutely, the ones you have already added are good. I am referring to https://github.com/ron-rs/ron/blob/0b7446179df381eaa215eb0eaa67f2ba68d1bca6/src/error.rs#L305 - could this encode also get a similar test?

sdroege: Ah! Sure :)

sdroege: Added something. I tried a lot to trigger this specific error, and that's the easiest I could find. If you have a better idea let me know.

juntyr:
LGTM, I'll give @torkleyy a few days to give a review as well, otherwise I'll merge :)

juntyr (approved):
LGTM

a1phyr:
As a RON user, I would prefer if `ron` sticked to use `base64`. It is a more widely used crate, and is far more likely to already be part of a dependency tree. A new dependency is not worth a few cosmetic change in this codebase IMO (I am not the one who maintains it, though).

juntyr:
> As a RON user, I would prefer if `ron` sticked to use `base64`. It is a more widely used crate, and is far more likely to already be part of a dependency tree. A new dependency is not worth a few cosmetic change in this codebase IMO (I am not the one who maintains it, though).

Thank you for voicing your concerns. I am pretty personally quite neural about which of the two libraries are used. `base64` has a larger user base, though `data-encoding` is no small library either. I'm open to listening to more feedback on this change and postpone merging until then.

juntyr:
Since RON will most likely be moving away from base64 (see #438), I will close this PR for now. Thank you @sdroege for your contribution!

PR #438, opened 2023-02-06
Rusty byte strings in RON, deprecate base64 (byte) strings

### LINKED ISSUE #436: RON incorrectly adds extra level of Base64 when roundtripping Bytes
This code ([playground]):
```rust
#[derive(Debug, serde::Deserialize, serde::Serialize)]
#[serde(rename = "b")]
struct BytesVal {
    pub b: bytes::Bytes,
}

#[derive(Debug, serde::Deserialize, serde::Serialize)]
#[serde(untagged)]
enum Bad {
    Bytes(BytesVal),
}

fn main() {
    let v: Bad = ron::from_str(r#"(b: "dGVzdA==")"#).unwrap();
    dbg!(&v);
    let s = ron::to_string(&v).unwrap();
    dbg!(&s);
    let v: Bad = ron::from_str(&s).unwrap();
    dbg!(&v);
    println!("---");
    let v: Bad = serde_json::from_str(r#"{"b": "dGVzdA=="}"#).unwrap();
    dbg!(&v);
    let s = serde_json::to_string(&v).unwrap();
    dbg!(&s);
    let v: Bad = serde_json::from_str(&s).unwrap();
    dbg!(&v);
}
```

has the following output:

```
[src/main.rs:23] &v = Bytes(
    BytesVal {
        b: b"dGVzdA==",
    },
)
[src/main.rs:25] &s = "(b:\"ZEdWemRBPT0=\")"
[src/main.rs:27] &v = Bytes(
    BytesVal {
        b: b"ZEdWemRBPT0=",
    },
)
---
[src/main.rs:30] &v = Bytes(
    BytesVal {
        b: b"dGVzdA==",
    },
)
[src/main.rs:32] &s = "{\"b\":[100,71,86,122,100,65,61,61]}"
[src/main.rs:34] &v = Bytes(
    BytesVal {
        b: b"dGVzdA==",
    },
)
```

It fully survives the roundtrip with `serde_json`, but adds extra level of Base64 with RON.

[playground]: https://www.rustexplorer.com/b/6yq8nt

juntyr:
# Breaking: Rusty byte strings in RON, deprecate base64 (byte) strings

ron has for now encoded byte-slices and byte-bufs as base64-encoded strings. While this works perfectly fine when ron deserialises a document using type hints (i.e. when [`Deserializer::deserialize_byte_buf`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_byte_buf) or [`Deserializer::deserialize_bytes`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_bytes) are called), #436 has shown that it breaks down when ron is asked to deserialise bytes in a self-describing document. This issue occurs when [`Deserializer::deserialize_any`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_any) is called, for instance, if the byte string is inside a `#[serde(untagged)]` enum.

This issue presents an opportunity to transition ron from using ambiguous base64 (byte) strings to using Rusty byte strings. For instance, the Rust value `b"Hello ron!"` is now encoded as the equivalent ron value `b"Hello ron!"` (instead of `"SGVsbG8gcm9uIQ=="` as before). ron's new Rusty byte strings follow the updated escape rules for Rust's byte strings from https://github.com/rust-lang/rfcs/pull/3349 (assuming that the RFC will be accepted).

## Request for Comment

As this is a breaking change in ron's format, that will affect user's of ron over a transition from v0.9 to v0.10 (see below), we would like to hear your thoughts on the transition and how to make it easier on you before we merge this PR.

### Serialising

In the next breaking release of ron, v0.9, all calls to [`Serializer::serialize_bytes`](https://docs.rs/serde/latest/serde/ser/trait.Serializer.html#tymethod.serialize_bytes) will output Rusty byte strings instead of base64-encoded strings.

### Deserialising

In the next breaking release of ron, v0.9, all calls to [`Deserializer::deserialize_byte_buf`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_byte_buf) and [`Deserializer::deserialize_bytes`](https://docs.rs/serde/latest/serde/trait.Deserializer.html#tymethod.deserialize_bytes) will accept both Rusty byte strings and base64-encoded (byte) strings. ron's [`Error::Base64Error`](https://docs.rs/ron/latest/ron/error/enum.Error.html#variant.Base64Error) variant will be deprecated, and will be removed in v0.10.

### Value

ron's `Value` type gains a new `Value::Bytes(Vec<u8>)` variant to store an owned byte buffer.

## What changes in v0.10?

One breaking release after the introduction of Rusty byte strings, base64-encoded (byte) strings will no longer be accepted during deserialising. The [`Error::Base64Error`](https://docs.rs/ron/latest/ron/error/enum.Error.html#variant.Base64Error) variant will be removed. However, ron *may* still produce non-generic error messages when deserialising a byte string and coming across a normal string that *looks* like it *may* be base64-encoded.

## How can I continue using base64-encoded (byte) strings?

If you would like to retain the original behaviour and continue to use (now strongly typed) base64 strings, you will need to change the [`Serialize`](https://docs.rs/serde/latest/serde/ser/trait.Serialize) implementation of your byte value, e.g. using one of the following crates (non-exhaustive and without endorsement from ron):

- [`serde_with`](https://crates.io/crates/serde_with): [`base64`](https://docs.rs/serde_with/latest/serde_with/base64/index.html) module contains several helper types, which can be used in a, e.g., `#[serde_as(as = "Base64")] v: Vec<u8>` annotation, uses a custom base64 implementation
- [`base64-serde`](https://docs.rs/base64-serde/latest/base64_serde/macro.base64_serde_type.html): [`base64_serde_type`](https://docs.rs/base64-serde/0.7.0/base64_serde/macro.base64_serde_type.html) macro generates a new wrapper type, uses [`base64`](https://crates.io/crates/base64)
- [`serde_bytes_base64`](https://crates.io/crates/serde_bytes_base64): [`Bytes`](https://docs.rs/serde_bytes_base64/latest/serde_bytes_base64/struct.Bytes.html) wrapper type, uses [`base64`](https://crates.io/crates/base64)

With this PR, we have also provided a new example ([`examples/base64.rs`](https://github.com/ron-rs/ron/blob/rusty-byte-str/examples/base64.rs)) of continuing to use base64-encoded bytes with ron. In this example, two `serde` helpers are implemented which showcase

1. how to serialise and deserialise base64-encoded (byte) strings in a strongly typed way that avoids #436
2. how to accept both Rusty byte strings and the legacy base64-encoded (byte) strings during deserialising

## Final notes

Fixes #436

## TODOs

* [x] Update the grammar
* [x] Document the extension
* [x] Add tests for the new `Value` behaviour
* [x] Add a test to document the v0.10 deprecation error while deserialising
* [x] Add tests to document the v0.9 optional base64 deserialising
* [x] Add a test to document a datatype-side workaround for only base64
* [x] Add a test to document a datatype-side workaround for base64 or Rusty byte strings
* [x] Add more code coverage tests for `parse.rs`, `error.rs` coverage can be handled separately
* [x] Change the example link to the main repo before merging


juntyr:
I still think that this is the right direction for RON to evolve in, as it makes the format more Rusty. However, the transition from old to new RON byte strings will be important to get right. Perhaps the transition could be eased in the following way:

1. base64 is only deprecated in v0.11, so in v0.9 and v0.10 both formats are allowed (but the feature gate will still be called "deprecated_base64_byte_string")
   * in v0.9 serialize will produce Rusty byte strings, and the `deprecated_base64_byte_string` extension will be enabled by default for deserialising
   * in v0.10 the `deprecated_base64_byte_string` extension is no longer enabled by default for deserialising
2. deserialising a byte string from a base64 string with type hints should be detected for one version longer than producing base64 byte strings (by checking for base64 and producing an informative error)
3. the expected byte string but found base64 error should produce a structured replacement suggestion
4. how should deserialising a string in deserialise_any that looks like base64 be handled?

src/parse.rs
```diff
@@ -877,16 +991,22 @@ impl<'a> Bytes<'a> {
...
+            b'\\' => EscapeCharacter::Ascii(b'\\'),
+            b'n' => EscapeCharacter::Ascii(b'\n'),
+            b'r' => EscapeCharacter::Ascii(b'\r'),
+            b't' => EscapeCharacter::Ascii(b'\t'),
+            b'0' => EscapeCharacter::Ascii(b'\0'),
+            b'x' => {
+                let b = self.decode_ascii_escape()?;
+                match encoding {
+                    EscapeEncoding::Binary => EscapeCharacter::Ascii(b),
+                    EscapeEncoding::Utf8 => EscapeCharacter::Utf8(b as char),
```
juntyr, line 1007: This is wrong: `\x..` may not be a valid UTF8 character by itself - we need to parse further bytes to find out if a valid character is found or not

bestouff:
I don't think you should ever remove the base64 deserialization code. Some people probably have RON files stored somewhere that they can't easily modify, and they need them to work.

juntyr:
Since no additional concerns have been voiced so far, I will merge this PR over the coming days.

@bestouff I would like to ensure that your concern about backwards compatibility is sufficiently resolved before base64 deserialisation is fully disabled (in v0.10 or later if necessary). Do you think that one of the above-described workarounds of moving the base64 encoding to the data type definition would work in your case?

PR #441, opened 2023-03-13
depend on `serde_derive` separately for more compilation parallelism

jakobhellermann:

If no crate enables the `derive` feature of `serde`, then `serde` and `serde_derive` can compile in parallel, and anything depending only on `serde` can compile while `serde_derive` is not yet done.
This can save up to 5 seconds of compilation, depending on whether this helps the critical chain and if no other crate enables the feature. 

image


juntyr (approved):
LGTM

PR #443, opened 2023-03-13
Upgrade to bitflags 2.0

juntyr:
Supersedes #442. Updates the `bitflags` dependency to 2.0

Breaking change:
- `bitflags` 2.0 serializes differently, now using the names instead of `(bits: 42)`. Did we ever commit to how `Extensions` is serialized?

@kvark [`bitflags-serial`](https://github.com/kvark/bitflags-serial) still uses `bitflags` 1.0, so I've removed the related test for now. Feel free to re-add it or ping me once `bitflags-serial` is updated


juntyr:
?r @torkleyy

PR #444, opened 2023-03-25
Support unicode idents (matching rust)

### LINKED ISSUE #321: Deserializer doesn't support unicode identifiers
```rust
use ron;

#[derive(Debug, Serialize, Deserialize, PartialEq)]
struct A {
    한글: String,
}

#[derive(Debug, Serialize, Deserialize, PartialEq)]
struct B {
    eng: String,
}
fn main() {
    let a = A {
        한글: "스트링".to_string(),
    };
    let b = B {
        eng: "string".to_string(),
    };
    let a_ser = ron::ser::to_string(&a).unwrap();
    let b_ser = ron::ser::to_string(&b).unwrap();
    let a_de = ron::de::from_str::<A>(&a_ser);
    let b_de = ron::de::from_str::<B>(&b_ser);
    println!("a_ser:\n{:#?}", a_ser );
    println!("a_de:\n{:#?}", a_de );
    println!("b_ser:\n{:#?}", b_ser );
    println!("b_de:\n{:#?}", b_de );
}
```

```console
a_ser:
"(한글:\"스트링\")"
a_de:
Err(
    Error {
        code: ExpectedIdentifier,
        position: Position {
            line: 1,
            col: 2,
        },
    },
)
b_ser:
"(eng:\"string\")"
b_de:
Ok(
    B {
        eng: "string",
    },
)
```

Is this a bug of ron?

ModProg:

This not only changes the ident parsing, but also uses `&str` instead of `&[u8]`
for the rest of the parsing (sometime as a byte slice through `.bytes()`).

If you prefer to have a more restricted implementation that only changes
`.*ident*()` methods on `Bytes` I can revert the other changes and use
`from_utf8_unchecked` instead.

But as I saw that the benchmark using this implementation was more or less on
par with the current implementation I wanted to share it at least:

`ron-new` is this PR's string based implementation, `ron` is the current git
version.

```
Benchmarking Serde Deserialization/json/data/canada: Warming up for 3.0000 s
Warning: Unable to complete 10000 samples in 5.0s. You may wish to increase target time to 89.8s, or reduce sample count to 550.
Serde Deserialization/json/data/canada
                        time:   [8.9827 ms 8.9938 ms 9.0056 ms]
                        change: [-3.4193% -2.9076% -2.4310%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 860 outliers among 10000 measurements (8.60%)
  1 (0.01%) low mild
  493 (4.93%) high mild
  366 (3.66%) high severe
Benchmarking Serde Deserialization/ron/data/canada: Warming up for 3.0000 s
Warning: Unable to complete 10000 samples in 5.0s. You may wish to increase target time to 317.2s, or reduce sample count to 150.
Serde Deserialization/ron/data/canada
                        time:   [31.000 ms 31.028 ms 31.056 ms]
                        change: [-1.5615% -1.2462% -0.9532%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 521 outliers among 10000 measurements (5.21%)
  209 (2.09%) high mild
  312 (3.12%) high severe
Benchmarking Serde Deserialization/ron-new/data/canada: Warming up for 3.0000 s
Warning: Unable to complete 10000 samples in 5.0s. You may wish to increase target time to 320.1s, or reduce sample count to 150.
Serde Deserialization/ron-new/data/canada
                        time:   [30.619 ms 30.638 ms 30.658 ms]
                        change: [-2.9465% -2.7434% -2.5733%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 164 outliers among 10000 measurements (1.64%)
  63 (0.63%) high mild
  101 (1.01%) high severe
```

I also didn't add any tests yet.

fixes #321

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -242,7 +242,7 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             return self.handle_any_struct(visitor);
         }
 
-        match self.bytes.peek_or_eof()? {
+        match self.bytes.peek_byte_or_eof()? {
             b'(' => self.handle_any_struct(visitor),
             b'[' => self.deserialize_seq(visitor),
             b'{' => self.deserialize_map(visitor),
@@ -686,7 +686,7 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
 
         match (
             self.had_comma,
-            self.de.bytes.peek_or_eof()? != self.terminator,
+            self.de.bytes.peek_byte_or_eof()? != self.terminator,
         ) {
             // Trailing comma, maybe has a next element
             (true, has_element) => Ok(has_element),
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -142,7 +142,7 @@ impl<'de> Deserializer<'de> {
         V: Visitor<'de>,
     {
         // Create a working copy
-        let mut bytes = self.parser;
+        let mut bytes = self.parser.clone();
 
         if bytes.consume_str("(") {
             bytes.skip_ws()?;
@@ -237,10 +237,7 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             return visitor.visit_f64(std::f64::NAN);
         }
 
-        // `identifier` does not change state if it fails
-        let ident = self.parser.identifier().ok();
-
-        if ident.is_some() {
+        if self.parser.skip_ident() {
             self.parser.skip_ws()?;
 
             return self.handle_any_struct(visitor);
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -63,7 +63,7 @@ impl<'de> Deserializer<'de> {
 
     pub fn remainder(&self) -> Cow<'_, str> {
         // FIXME this does not make sense with the unicode validation on creation
-        String::from_utf8_lossy(self.parser.bytes())
+        Cow::Borrowed(self.parser.src())
     }
 
     pub fn span_error(&self, code: Error) -> SpannedError {
@@ -125,7 +125,7 @@ impl<'de> Deserializer<'de> {
     pub fn end(&mut self) -> Result<()> {
         self.parser.skip_ws()?;
 
-        if self.parser.bytes().is_empty() {
+        if dbg!(self.parser).src().is_empty() {
             Ok(())
         } else {
             Err(Error::TrailingCharacters)
@@ -480,14 +480,13 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         V: Visitor<'de>,
     {
         if name == crate::value::raw::RAW_VALUE_TOKEN {
-            let bytes_before = self.parser.bytes();
+            let src_before = self.parser.src();
             self.parser.skip_ws()?;
             let _ignored = self.deserialize_ignored_any(serde::de::IgnoredAny)?;
             self.parser.skip_ws()?;
-            let bytes_after = self.parser.bytes();
+            let src_after = self.parser.src();
 
-            let ron_bytes = &bytes_before[..bytes_before.len() - bytes_after.len()];
-            let ron_str = str::from_utf8(ron_bytes).map_err(Error::from)?;
+            let ron_str = &src_before[..src_before.len() - src_after.len()];
 
             return visitor
                 .visit_borrowed_str::<Error>(ron_str)
@@ -688,10 +687,7 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
     fn has_element(&mut self) -> Result<bool> {
         self.de.parser.skip_ws()?;
 
-        match (
-            self.had_comma,
-            self.de.parser.peek()? != self.terminator,
-        ) {
+        match (self.had_comma, self.de.parser.peek()? != self.terminator) {
             // Trailing comma, maybe has a next element
             (true, has_element) => Ok(has_element),
             // No trailing comma but terminator
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -125,7 +125,7 @@ impl<'de> Deserializer<'de> {
     pub fn end(&mut self) -> Result<()> {
         self.parser.skip_ws()?;
 
-        if dbg!(self.parser).src().is_empty() {
+        if self.parser.src().is_empty() {
             Ok(())
         } else {
             Err(Error::TrailingCharacters)
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -143,7 +143,7 @@ impl<'de> Deserializer<'de> {
         // Create a working copy
         let mut bytes = self.parser.clone();
 
-        if bytes.consume_str("(") {
+        if bytes.consume_char('(') {
             bytes.skip_ws()?;
 
             if bytes.check_tuple_struct()? {
@@ -173,7 +173,7 @@ impl<'de> Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
-        if self.newtype_variant || self.parser.consume_str("(") {
+        if self.newtype_variant || self.parser.consume_char('(') {
             let old_newtype_variant = self.newtype_variant;
             self.newtype_variant = false;
 
@@ -194,7 +194,7 @@ impl<'de> Deserializer<'de> {
 
             self.parser.skip_ws()?;
 
-            if old_newtype_variant || self.parser.consume_str(")") {
+            if old_newtype_variant || self.parser.consume_char(')') {
                 Ok(value)
             } else {
                 Err(Error::ExpectedStructLikeEnd)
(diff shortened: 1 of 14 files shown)


src/parse.rs
```diff
@@ -111,7 +112,8 @@ pub enum AnyNum {
 pub struct Bytes<'a> {
     /// Bits set according to the [`Extensions`] enum.
     pub exts: Extensions,
-    bytes: &'a [u8],
+    // bytes: &'a [u8],
```
juntyr, line 115: nit: this should be removed

src/parse.rs
```diff
@@ -126,9 +128,13 @@ pub(crate) type LargeSInt = i64;
 
 impl<'a> Bytes<'a> {
     pub fn new(bytes: &'a [u8]) -> SpannedResult<Self> {
+        // // Verify that input is valid utf-8
```
juntyr, line 131: seems like some leftover code

juntyr:
@ModProg I've only had time to briefly skim your changes so far, but (formatting excluded) they look good so far. Adding some tests would definitely be the next step, and I'll try to review the PR more thoroughly over the coming days. Thank you already for your work!

ModProg:
Yes, def needs some cleanup.

juntyr:
Once you get through the final cleanup, there are also still some `dbg!()` macros left

juntyr:
Could you also add tests for the new crash cases you discovered to document which errors they should produce (and on which column)?

src/parse.rs
```diff
@@ -111,7 +112,7 @@ pub enum AnyNum {
 pub struct Bytes<'a> {
```
ModProg, line 112: Maybe `Bytes` isn't the best name anymore

juntyr: That’s true - what about Parser?

src/parse.rs
```diff
@@ -500,10 +518,14 @@ impl<'a> Bytes<'a> {
...
+            let _ = self.advance_byte();
 
-        Ok(peek)
+            Ok(peek)
+        } else {
+            Err(Error::NoCharBoundary)
+        }
     }
 
     pub fn expect_byte(&mut self, byte: u8, error: Error) -> Result<()> {
```
juntyr, line 531: Do we still actually need this method? Would it be possible to eat (and expect) `char`s instead?

src/parse.rs
```diff
@@ -73,16 +74,16 @@ const fn is_float_char(c: u8) -> bool {
...
 
-pub const fn is_ident_other_char(c: u8) -> bool {
-    ENCODINGS[c as usize] & IDENT_OTHER_CHAR != 0
-}
-
-pub const fn is_ident_raw_char(c: u8) -> bool {
-    ENCODINGS[c as usize] & IDENT_RAW_CHAR != 0
+pub fn is_ident_raw_char(c: char) -> bool {
+    if c.is_ascii() {
+        ENCODINGS[c as usize] & IDENT_RAW_CHAR != 0
```
juntyr, line 83: Which of the initial ascii byte classes for the encoding table are still necessary?

src/parse.rs
```diff
@@ -676,7 +705,7 @@ impl<'a> Bytes<'a> {
     pub fn skip_ws(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
ModProg: yes, that sounds like a good idea.

src/parse.rs
ModProg: We probably don't, it might pose a marginal performance benefit, but probably doesn't matter.

src/parse.rs
ModProg: we can probably shrink that a bit. I was also thinking about doing a performance comparison with std functions like [`is_digit`](https://doc.rust-lang.org/std/primitive.char.html#method.is_digit)

> I am wondering if we could not entirely stop relying on bytes but replace all "next byte" methods with "next character" methods. Would this be possible? What are your thoughts @ModProg?

ModProg:
> I am wondering if we could not entirely stop relying on bytes but replace all "next byte" methods with "next character" methods. Would this be possible? What are your thoughts @ModProg?

probably, we should run the benchmark with that change

ModProg:
I just noticed that I made a stupid mistake while benchmarking and benchmarked the wrong versions... The performance impact of some of my changes is larger than I thought and I will have to check that tomorrow.

ModProg:
> I just noticed that I made a stupid mistake while benchmarking and benchmarked the wrong versions... The performance impact of some of my changes is larger than I thought and I will have to check that tomorrow.

(in total they ended up increasing the benchmark by ca. 50%)

src/de/mod.rs
```diff
@@ -42,28 +42,32 @@ impl<'de> Deserializer<'de> {
...
+    pub fn from_bytes_with_options(input: &'de [u8], options: Options) -> SpannedResult<Self> {
+        Self::from_str_with_options(
+            str::from_utf8(input).map_err(|error| SpannedError::from_utf8_error(error, input))?,
+            options,
+        )
+    }
+
     pub fn remainder(&self) -> Cow<'_, str> {
-        String::from_utf8_lossy(self.bytes.bytes())
+        // FIXME this does not make sense with the unicode validation on creation
```
juntyr, line 65: While we could probably break the API here by always returning the remaining `&str`, it may be easier to give us more future wiggle room for now and just always return a borrowed `Cow` that just contains the remaining str

juntyr:
> > I just noticed that I made a stupid mistake while benchmarking and benchmarked the wrong versions... The performance impact of some of my changes is larger than I thought and I will have to check that tomorrow.
> 
> (in total they ended up increasing the benchmark by ca. 50%)

That is significant ... perhaps once the code changes have stabilised a bit we can look into where the perf hit comes from. So far your changes seem to make the code more readable, which would be a counter-benefit.

src/parse.rs
```diff
@@ -168,34 +170,125 @@ impl<'a> Bytes<'a> {
...
-    pub fn advance_char(&mut self) -> Result<usize> {
-        let c = self.peek_char_or_eof()?;
-        if c == '\n' {
-            self.cursor.line += 1;
-            self.cursor.col = 1;
+    pub fn peek(&self) -> Result<char> {
+        self.src.chars().next().ok_or(Error::Eof)
+    }
+
+    pub fn bytes(&self) -> &'a [u8] {
```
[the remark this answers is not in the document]
ModProg: `escaped_string`, and the start of both `string` and `integer`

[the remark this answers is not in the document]
juntyr: I think what I was wondering about is if there are any cases left where we need to check bytes instead of chars for functionality. Do you think code that only relied on chars would be cleaner?

ModProg:
@juntyr  I got performance back to where it was before this PR

juntyr:
> @juntyr I got performance back to where it was before this PR, but it would maybe be best if someone could verify.

Ok, that sounds wonderful! I'll have a deeper look sometime later, check the performance, and see if I can find anything else :)

juntyr:
@ModProg Ok, I've been doing a bit of toying around with the changes. First, I just did a bit of cleanup with clippy and extended the docs. I also tested removing all possible byte strings from parsing, please feel free to disregard those changes. You can find my experiments here:
https://github.com/juntyr/ron/commit/c7e428bcf52c11be3ec3d52457b932e7fd4ab8f2

I also tested the other side of ron, serialisation. Previously we also used byte strings heavily there and just hoped that it would all come out as UTF8 on the other end. I've tested switching it to UTF8 without breaking the existing API surface in:
https://github.com/juntyr/ron/commit/c6c0997d8a1fc235f5198b4f20b605b8347f8378

I'm quite happy with this PR! If you include the clippy fixes, we can definitely land the existing changes and add onto them with UTF8 serialisation and a better `Value` for ints in follow-up commits.

ModProg:
> I also tested removing all possible byte strings from parsing, please feel free to disregard those changes. You can find my experiments here:
> [juntyr@c7e428b](https://github.com/juntyr/ron/commit/c7e428bcf52c11be3ec3d52457b932e7fd4ab8f2)

Those seem to add about 30% of time to the benchmark on my machine (from 29 to 38ms)

juntyr:
(sorry for the long review delay, I'm finishing up my second thesis atm and it will occupy me for a few more days)

ModProg:
> (sorry for the long review delay, I'm finishing up my second thesis atm and it will occupy me for a few more days)

no worries, I am in no rush.

juntyr:
@ModProg I am really sorry that it has taken me so long to get back to this PR, life has been happening.

I want to get this PR in before v0.9, as it is the last big change I intend to include in that. I will get started with rebasing the PR (on top of #438) over the coming days and will try to get it merged soon.

I am slowly trying to make ron even more Rusty (e.g. with Rusty byte strings in #438 and typed number suffixes and underscores in floats in #481), and this PR is an important step in that direction. Thank you so much for working on it!

juntyr:
#488 has now landed

Issue #357, opened 2022-01-04
Round-trip failure: untagged enum containing Vec variant is improperly deserialized


obi1kenobi:
Consider the following types:
```rust
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
#[serde(untagged)]
enum MyValue {
    Int(i64),
    String(String),
    Enum(Enum),
    List(Vec<MyValue>),
}

#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
enum Enum {
    First(String),
    Second(i64),
}
```

Let's try serializing a `MyValue::Enum` value like the following: `MyValue::Enum(Enum::First("foo".to_string()))`. Its RON-serialized value is `First("foo")`.

That `First("foo")` value unfortunately deserializes to `MyValue::List([MyValue::String("foo")])` instead of our original enum-based value. This is surprising to me, because the deserialization code decides to ignore the `First` prefix entirely, and skip forward until the `(` then deserialize a list with it. It's especially surprising because the `MyValue::Enum` variant is before the `MyValue::List` variant, and yet we still get a `MyValue::List` back instead of `MyValue::Enum`.

Here's a complete repro:
```rust
#[cfg(test)]
mod tests {
    use serde::{Serialize, Deserialize};

    #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
    #[serde(untagged)]
    enum MyValue {
        Int(i64),
        String(String),
        Enum(Enum),
        List(Vec<MyValue>),
    }

    #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
    enum Enum {
        First(String),
        Second(i64),
    }

    fn serialize_then_deserialize(value: &MyValue) -> MyValue {
        ron::from_str(ron::to_string(value).unwrap().as_str()).unwrap()
    }

    #[test]
    fn test() {
        let value = MyValue::Enum(Enum::First("foo".to_string()));
        let deserialized = serialize_then_deserialize(&value);
        assert_eq!(value, deserialized, "Serialized as: {}", ron::to_string(&value).unwrap());
    }
}
```

This fails with:
```
running 1 test
thread 'file::tests::test' panicked at 'assertion failed: `(left == right)`
  left: `Enum(First("foo"))`,
 right: `List([String("foo")])`: Serialized as: First("foo")', src/file.rs:X:9
```

Thanks to everyone that contributes to building and maintaining this awesome library! Any advice on avoiding this issue would be much appreciated.

[...]

PR #451, opened 2023-04-19
Add minimal support for internally tagged and untagged enums

### LINKED ISSUE #449: C-like enum nested inside of an internally tagged enum fails to deserialize
Trying to roundtrip this type Rust-RON-Rust:

```rust

#[derive(Debug, Serialize, Deserialize)]
enum InnerEnum {
    UnitVariant,
}

#[derive(Debug, Serialize, Deserialize)]
struct Container {
    field: InnerEnum,
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "tag")]
enum OuterEnum {  // <- this type
    Variant(Container),
}
```

gives the following RON:

```ron
(
    tag: "Variant",
    field: UnitVariant,
)
```

which produces the following error when trying to deserialize it:

```
4:2: Expected string or map but found a unit value instead
```

([playground]-ish)

Some notes:
1. Deserialization to `ron::Value` reveals that the variant of the `InnerEnum` is lost and replaced with `Unit`.
2. The error location is particularly unhelpful: it points to the end of the file.

<details><summary>Full repro code duplicated here for resillience</summary>

```rust
/*
[dependencies]
serde = { version = "1", features = ["derive"] }
ron = "=0.8.0"
*/

#[derive(Debug, Serialize, Deserialize)]
enum InnerEnum {
    UnitVariant,
}

#[derive(Debug, Serialize, Deserialize)]
struct Container {
    field: InnerEnum,
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "tag")]
enum OuterEnum {
    Variant(Container),
}

fn main() -> Result<(), ron::Error> {
    let value = OuterEnum::Variant(Container {
        field: InnerEnum::UnitVariant,
    });
    let serialized = ron::ser::to_string_pretty(&value, <_>::default())?;
    println!("--- serialized ---\n{serialized}\n");
    let ron_value: ron::Value = ron::from_str(&serialized)?;
    println!("--- deserialized as ron::Value ---\n{ron_value:#?}\n");
    let err = ron::from_str::<OuterEnum>(&serialized).unwrap_err();
    println!("--- deserialization error ---\n{err}");

    Ok(())
}
```

</details>

Possibly related to #217, but doesn’t use neither untagged nor externally tagged enums.

[playground]: https://www.rustexplorer.com/b/k43br1

juntyr:
Quite hacky but minimally invasive implementation to support internally tagged and untagged enums.

This PR switches the self-describing enum deserialisation behaviour to match serde's internal JSONy `Content` type format if we think we are deserialising into serde's internal type. Specifically, it uses `std::any::type_name` to detect when `serde::__private::de::content::Content` is deserialised to. This is relying on undocumented behaviour in all the horrible ways ... but that's what we need anyways need to do if we want to support internally tagged and untagged enums in serde. In some distant future, https://github.com/serde-rs/serde/pull/2420 might make this less horrible.

Unlike #409, this PR is very targeted as it does not change the deserialisation behaviour for any other types. This allows us to stick with the existing self-describing enum behaviour and gives us space to evolve `ron::Value` separately.

Closes #449


juntyr:
@torkleyy I really don't like that serde forces us to do this but the issue comes up again and again and a slightly hacky solution that relies on observable serde behaviour that has not changed since 1.0.0 (https://github.com/serde-rs/serde/blame/0c6a2bbf794abe966a4763f5b7ff23acb535eb7f/serde/src/private/de.rs#L220) and does not change RON's behaviour in any other way might be the best we can get for the next few years. I'd thus suggest merging this (with a heavy heart).

juntyr:
Dang, now I remember why #409 got so large ... Unfortunately supporting untagged and internally tagged enums also means needing to support self-describing newtype variants for these cases ... I also cleaned up our tuple struct detection along the way but it still becomes a bit like spaghetti code since something newtype-looking cannot just always be unwrapped since our `ron::Value` still depends on these structs to become singleton tuples ... the code may become slightly cleaner if I also introduce better support for `Value` so that it just supports newtypes itself, but for now it's the best I can do

tests/117_untagged_tuple_variant.rs
```diff
@@ -23,7 +23,7 @@ fn test_ebkalderon_case() {
     flags: [
         "--enable-thing",
         "--enable-other-thing",
-        If("some-conditional", ["--enable-third-thing"]),
```
torkleyy, line 26: Why did you remove this?

torkleyy (commented):
This seems a like a decent approach given our options 👍 
I think we'll need some documentation. Unfortunate that these hacks are needed...

tests/117_untagged_tuple_variant.rs
juntyr: I have no idea why this worked before, since we are testing untagged enums here. I think by adding the `If` it previously worked despite us missing support for untagged enums. However, now that it is properly implemented, you cannot give the name of the variant since untagged enums do not expect that

juntyr:
TODO @juntyr: check if #357 would be solved by this

juntyr:
> TODO: check if #357 would be solved by this

#477 shows that #357 was indeed fixed by this

PR #453, opened 2023-04-23
Add minimal support for flattening roundtrips through maps

### LINKED ISSUE #115: Struct flattening not working

juntyr:
Alternative to #403 which adds minimal support for `#[serde(flatten)]` to RON. Instead of many workarounds to make flattened structs serialise into nice RON struct syntax and deserialise from them, both of which are really difficult since serde gives us no information to work with and just delegates to map serialisation/deserialisation, this PR just ensures that at least flattened structs can roundtrip through RON.

For this, just one hack is necessary. Map keys that request identifiers can now be deserialised from strings, sort of. But only if they are written exactly as `"`, identifier, `"`. This seems to be enough to get the tests to work that I initially designed for #403.

@torkleyy I'd be in favour of landing this instead of going down the rabbit hole of #403 since that just fixes one of the two parts and getting flattened structs to serialise correctly would require hacks of another order.

This would not really fix #115 but at least get some working if imperfect solution.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -26,6 +26,7 @@ mod value;
 pub struct Deserializer<'de> {
     bytes: Bytes<'de>,
     newtype_variant: bool,
+    flatten_map_key: bool,
     last_identifier: Option<&'de str>,
     recursion_limit: Option<usize>,
 }
@@ -49,6 +50,7 @@ impl<'de> Deserializer<'de> {
         let mut deserializer = Deserializer {
             bytes: Bytes::new(input)?,
             newtype_variant: false,
+            flatten_map_key: false,
             last_identifier: None,
             recursion_limit: options.recursion_limit,
         };
@@ -420,6 +422,8 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
+        self.flatten_map_key = false;
+
         if self.bytes.consume("None") {
             visitor.visit_none()
         } else if self.bytes.consume("Some") && {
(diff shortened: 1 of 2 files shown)


clouds56:
you could cherry-pick #454 here?

juntyr:
> you could cherry-pick #454 here?

I could if you prefer that. However, we could also land #454 with some minor changes :)

clouds56:
Sure, just go ahead.
I'm hurt by serde and need some rest.

juntyr:
> Sure, just go ahead.
> I'm hurt by serde and need some rest.

Ok, I'll cherry pick your changes later today :)

juntyr:
Superseded by #455

PR #455, opened 2023-04-24
Add minimal support for flattening roundtrips through maps

juntyr:
Supersedes #403, #453, and #454 to add minimal support for `#[serde(flatten)]` in RON. Instead of many workarounds to make flattened structs serialise into nice RON struct syntax and deserialise from them, both of which are really difficult since serde gives us no information to work with and just delegates to map serialisation/deserialisation, this PR just ensures that at least flattened structs can roundtrip through RON.

For this, just one hack is necessary. If we detect that a map is deserialised into what looks like a struct (thanks @clouds56 for contributing the detection in #454), we enforce that map key identifiers are actually strings written exactly as `"`, identifier, `"`. This seems to be enough to get the tests to work that I initially designed for https://github.com/ron-rs/ron/pull/403.

@torkleyy I'd be in favour of landing this instead of going down the rabbit hole of https://github.com/ron-rs/ron/pull/403 since that just fixes one of the two parts and getting flattened structs to serialise correctly would require hacks of another order.

This would not really fix https://github.com/ron-rs/ron/issues/115 but at least get some working if imperfect solution.

* [x] Add tests for the new error cases, i.e. when flattening goes wrong

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -601,10 +601,9 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
 
         self.newtype_variant = false;
 
-        let mut buffer = [0u8; SERDE_FLATTEN_CANARY.len()];
-        let mut cursor = std::io::Cursor::new(&mut buffer as &mut [u8]);
-        let _ = write!(cursor, "{}", VisitorExpecting(&visitor));
-        let terminator = if buffer == SERDE_FLATTEN_CANARY {
+        let mut canary_buffer = [0u8; SERDE_FLATTEN_CANARY.len()];
+        let _ = write!(canary_buffer.as_mut(), "{}", VisitorExpecting(&visitor));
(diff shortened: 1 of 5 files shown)


src/de/mod.rs
```diff
@@ -601,9 +601,10 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
 
         self.newtype_variant = false;
 
-        let mut cursor = std::io::Cursor::new([0u8; SERDE_FLATTEN_CANARY.len()]);
+        let mut buffer = [0u8; SERDE_FLATTEN_CANARY.len()];
+        let mut cursor = std::io::Cursor::new(&mut buffer as &mut [u8]);
         let _ = write!(cursor, "{}", VisitorExpecting(&visitor));
```
clouds56, line 606:
```rust
let _ = write!(buffer.as_mut(), ...);
```

juntyr: Thanks, I've applied your suggestion :)

PR #459, opened 2023-06-26
Bump indexmap to 2.0 and update MSRV to 1.64.0

juntyr:
The MSRV update is required by hashbrown 0.14 which indexmap 2.0 depends on.


juntyr:
?r @torkleyy

PR #461, opened 2023-07-02
Remove Cargo.lock for fuzzing

juntyr:


juntyr:
CI currently fails since grcov hasn't updated proc-macro2 in a while either - fix fuzzing first.

PR #466, opened 2023-07-18
Fix "WASD" example in readme

tje:
This may seem trivial, but when I noticed it I nearly screamed and called the police


juntyr:
Oh dang, good you saw that :D

Issue #241, opened 2020-05-26
Lossless primitive types


cart:
I'm using ron to deserialize types that are not known at compile time. It would be really nice to have the option to fully encode primitive types in ron so that type information isn't lost.

Ex: if i serialize `1u32`, it will be written as `1`. If I then deserialize it using deserialize_any, it will become `1u8` because it falls into the u8 range.

If instead `1u32` got serialized as `1u32`, this would not be an issue. It would also allow ron to skip the `any_num` range checks.

I propose a new `PrettyConfig::explicit_primitive_types` option that enables this behavior.

kvark:
That sounds quite reasonable to me 👍 
The only concern is that this is just one more step towards the format being self-describing, and we know that we aren't going to reach there. For example, you can't reconstruct an enum type efficiently just by looking at its variants.

cart:
Fortunately my use case doesn't require it to be fully self describing. This is the only change I want. It's just a small iterative improvement to an already working system.

kvark:
comment

juntyr:
I think especially with #479, I'm now convinced that this is a good idea to ensure that values can roundtrip. I'll see if I can work on this tomorrow.

As I'm on it, I should also check the integer and float parsing rules to ensure that both accept `_` in the correct places, such that number parsing should just be:
- identify the byte span of the number, as before
- check after this span for a number type suffix
- check if i128/u128 is allowed
- give a nice error if it's a float literal but with an int prefix
- give a nice error if it's a negative literal but an unsigned integer
- use existing parsing logic, but constrain the types to the specified ones

PR #479, opened 2023-08-19
Lossless `Value::Number` and allow +unsigned

### LINKED ISSUE #445: plus sign on unsigned integers
I don't see a reason why `+` should only be allowed on signed integers.

Currently this fails:
```rs
    assert_eq!(
        std::u64::MAX,
        from_str("+19").unwrap()
    );
```

juntyr:
This PR

* cleans up the `value` module a bit
* expands the `value::Number` enum to cover all possible number types losslessly
* merges the private `parser::AnyNum` type with `value::Number`
* fixes #445 by allowing `+unsigned` to parse as an unsigned integer

I see this as a small improvement towards a better `Value`. While it does touch a lot of tests, since `Number` is more explicit now and always chooses the smallest variant that fits (and `serde`'s `Deserialize` impls do the rest by upcasting again if necessary), more control could be given to the user with #241, if someone wants to implement that.


src/parse.rs
```diff
@@ -858,6 +816,8 @@ impl<'a> Bytes<'a> {
     }
 
     pub fn unsigned_integer<T: Num>(&mut self) -> Result<T> {
```
juntyr, line 818: Since `signed_integer` and `unsigned_integer` are still exposed, they should get extra tests with strong typing

PR #481, opened 2023-08-21
Add number suffixes and allow more number underscores

### LINKED ISSUE #241: Lossless primitive types

juntyr:
Fixes #241

This PR implements optional number type suffixes, which can be turned on with the `PrettyConfig::number_suffixes` option and are allowed during parsing. To allow Rusty suffixes to work with floating literals, this PR also finally allows using underscores in float literals, using the Rust syntax as a reference.

This PR turned out to be significantly more work than I had anticipated. In particular, a number which could fit into the destination type, e.g. `1_u8`, now has to reject when deserialising into a typed `u16`, since the type is an explicit mismatch. I tried my best to keep the changes in parsing small-ish and to use this opportunity to improve some error reporting a bit.

For users of RON, this should be an unobservable change (apart from some error code changes which are now more meaningful). However, writing RON files can now be a bit more Rusty as types can be encoded explicitly and underscores can be used to make (long) number literals more readable. However, anyone serialising RON data should probably not turn the `PrettyConfig::number_suffixes` flag on unless they need lossless roundtrips through `Value::Number` variants, since explicit typing makes it harder to upgrade a format to different number types without breaking compatibility.


juntyr:
Manually checking the coverage shows that 100% of the patch seems to be covered

src/ser/mod.rs
```diff
@@ -500,24 +540,36 @@ impl<W: io::Write> Serializer<W> {
...
     }
 
     fn write_identifier(&mut self, name: &str) -> Result<()> {
         self.validate_identifier(name)?;
         let mut bytes = name.as_bytes().iter().copied();
-        if !bytes.next().map_or(false, is_ident_first_char) || !bytes.all(is_ident_other_char) {
+        if !bytes.next().map_or(false, is_ident_first_char)
+            || !bytes.all(is_ident_other_char)
+            || name == "inf"
+            || name == "NaN"
```
juntyr, line 571: This should also check for `inf_f32`, `inf_f64`, `NaN_f32`, and `NaN_f64`.

src/parse.rs
```diff
@@ -184,70 +184,57 @@ impl<'a> Bytes<'a> {
...
-
-        let bytes_backup = self.bytes;
-
-        if self.next_bytes_is_float() {
-            return any_float(self.float::<f64>()?);
+            calc_num(self, &self.bytes[0..num_bytes], base, T::checked_sub_ext)
         }
+    }
 
+    pub fn any_integer(&mut self, desire: DesireInteger) -> Result<ParsedInteger> {
```
juntyr, line 237: Could this be refactored to again take a generic with a trait that takes care of implementing the parsing and unwrapping from a parsed int, same for float, this should allow moving the error handling to parse and using the literal in the error message

docs/grammar.md
```diff
@@ -47,16 +47,26 @@ value = integer | float | string | char | bool | option | list | map | tuple | s
...
+integer = ["+" | "-"], unsigned, [integer_suffix];
+integer_suffix = ("i", "u"), ("8", "16", "32", "64", "128");
+
+unsigned = unsigned_binary | unsigned_octal | unsigned_hexadecimal | unsigned_decimal;
+unsigned_binary = "0b", digit_binary, { digit_binary | "_" };
+unsigned_octal = "0o", digit_octal, { digit_octal | "_" };
+unsigned_hexadecimal = "0x", digit_hexadecimal, { digit_hexadecimal | "_" };
+unsigned_decimal = digit, { digit | "_" };
+
 float = ["+" | "-"], ("inf" | "NaN" | float_num);
```
juntyr, line 63: The float suffixed need to be in the top rule since they can apply to inf and NaN as well - also, is inf_f32 really the best? f32::inf would be an alternative (but inconsistent syntax)

juntyr:
?r @torkleyy and @cart

torkleyy (approved):

PR #483, opened 2023-08-22
Try slightly modified codecov CI

juntyr:
This PR improves our code coverage generation and moves our coverage data from codecov.io to self-hosting it on GitHub pages (since the reports on codecov.io have very weird lines whereas grcov's HTML report is much better). On PRs, the HTML reports should now be accessible as artefacts, on pushes to the main branch, the coverage data is re-deployed.


juntyr:
It seems this may be a codecov problem, see https://github.com/rust-lang/rust/issues/102477

PR #485, opened 2023-08-23
Fix value explosion in arbitrary fuzzer

juntyr:
Fix fuzzer-found issue [#61686](https://oss-fuzz.com/issue/6271570753224704)


juntyr:
I took some more time into further cases where generating a length from simply a usize might be problematic. Arrays are the big issue here since arbitrary has no idea that the array type, which only recurses into the O(1) kind, actually expands to O(N). I experimented with many approaches, but eventually settled on a hardcoded maximum span to ensure that deeply nested arrays don't explode the document. This approach is also reasonably performant during fuzzing.

PR #487, opened 2023-08-23
Fix raw value constructor vs de parsing incompatibility

juntyr:
* [x] Add test case for fuzzer-found hardcoded identifier bug


juntyr:
@torkleyy This is quite an annoying bug and I am not sure if this is the right way forward to fix it. The problem is that `RawValue` will have to be annoying in one of two ways:

a) `RawValue::from_ron("/*hi*/ 42 /*bye*/")` works as expected (i.e. any valid ron, except for attributes, can be used in the constructor and will be pasted literally into the output). However, deserialising only roundtrips if no pretty serialising is used since any pretty whitespace or comments will be eaten up by the raw value upon the next deserialising. I'm going with this option right now.

b) `RawValue::from_ron("/*hi*/ ( /*yay*/ 42 /*lol*/ ) /*bye*/")` only keeps `( /*yay*/ 42 /*lol*/ )` (so all whitespace before and after the first non-whitespace is lost), but roundtripping works since additional pretty whitespace is dropped.

Which one would you prefer?

PR #502, opened 2023-09-11
Fuzz serde enum representations

### LINKED ISSUE #500: Untagged enum roundtrip issues
The following cases do not roundtrip:

* untagged struct variant with `PrettyConfig::default().struct_names(true)`, since
  * serde serializes an untagged struct variant as a struct (ok)
  * the struct is given the name of the enum, which we dutifully serialize (oh no), see https://github.com/serde-rs/serde/blob/ddc1ee564b33aa584e5a66817aafb27c3265b212/serde_derive/src/ser.rs#L960-L964
  * the untagged enum deserializer does NOT allow any idents, since these are interpreted as variant names, which we interpret as enums, while serde is looking for a struct (dang)
  * I have no idea how to solve this, @clouds56 do you have any ideas?
* raw value inside an untagged enum
  * raw values are serialized without any indication that they exist, as is their purpose
  * untagged enums deserialize into serde's content type, which will parse the raw value's content as normal RON
  * the raw value cannot be deserialized from deserialized ron
  * e.g. `"\"hi\""` is the raw value `"\"hi\""` but the serde content `"hi"`
  * this could only be solved with a custom content type, which is blocked https://github.com/serde-rs/serde/issues/1183

juntyr:
Fixes #500, since all (known) roundtrip issues are now documented and tested. Once serde allows us to fix them, they can be removed from the newly expanded restrictions section.

- [x] externally tagged enums
- [x] internally tagged enums
- [x] internally tagged structs
- [x] adjacently tagged enums
- [x] untagged enums
- [x] other variant in internally tagged enums
- [x] other variant in adjacently tagged enums
- [x] serialize flattened struct
- [x] serialize flattened struct variant
- [x] deserialize flattened struct
- [x] deserialize flattened struct variant
  - [x] externally tagged
  - [x] untagged
  - [x] adjacently tagged
  - [x] internally tagged

* [x] document one-tuple/one-array limitation inside unwrapped newtype variants
* [x] ~~document that serde only supports unit, bool, integer, float, char, string, and byte keys inside flattened structs, and only string and byte keys and `u8` and `u64` when the flattened struct is inside an untagged or internally tagged enum~~ document that we only support string keys inside flattened maps (even though serde technically permits more, this support is not roundtrip-safe as soon as flattened maps are involved)
* [x] document that serde does not support flattened structs containing conflicting keys, e.g. an inner key matches a later outer key or two flattened maps in the same struct share a key
* [x] document that `#[enable(implicit_some)]` is not supported with `Option`s inside `#[serde(flatten)]`
* [x] document that `PrettyConfig::struct_names(true)` is not supported inside flattened structs / struct variants

- [x] Add tests for any bug fixes
- [x] Add tests for all known bugs
  - [x] structs or struct variants that contain a `#[serde(flatten)]`ed field: 
    - [x] are only serialised as maps and deserialised from maps
    - [x] must not contain duplicate fields / keys, e.g. where an inner-struct field matches an outer-struct or inner-struct field
    - [x] must not contain more than one (within the super-struct of all flattened structs) `#[serde(flatten)]`ed map field, which collects all unknown fields
    - [x] if they contain a `#[serde(flatten)]`ed map, they must not contain:
      - [x] a struct that is not flattened itself but contains some flattened fields and is flattened into the outer struct (variant)
      - [x] an untagged struct variant that contains some flattened fields
      - [x] a flattened externally tagged newtype, tuple, or struct variant, flattened internally tagged unit, newtype, or struct variant, or any flattened adjacently tagged variant
      - [x] a flattened tagged struct
  - [x] ron only supports string keys inside maps flattened into structs
  - [x] internally (or adjacently) tagged or untagged enum variants or `#[serde(flatten)]`ed fields must not contain:
    - [x] `i128` or `u128` values
    - [x] struct names, e.g. by enabling the `PrettyConfig::struct_names` setting
    - [x] newtypes
    - [x] zero-length arrays / tuples / tuple structs / structs / tuple variants / struct variants
      - [x] `Option`s with `#[enable(implicit_some)]` must not contain any of these or a unit, unit struct, or an untagged unit variant
    - [x] externally tagged tuple variants with just one field (that are not newtype variants)
    - [x] tuples or arrays with just one element are not supported inside newtype variants with `#[enable(unwrap_variant_newtypes)]`
    - [x] a `ron::value::RawValue`
  - [x] untagged tuple / struct variants with no fields are not supported
  - [x] untagged tuple variants with just one field (that are not newtype variants) are not supported when the `#![enable(unwrap_variant_newtypes)]` extension is enabled
  - [x] internally tagged newtype variants and `#[serde(flatten)]`ed fields must not contain:
    - [x] a unit or a unit struct inside an untagged newtype variant
    - [x] an untagged unit variant
  - [x] serializing a `ron::value::RawValue` using a `PrettyConfig` may add leading and trailing whitespace and comments, which the `ron::value::RawValue` absorbs upon deserialization
- [x] I've included my change in `CHANGELOG.md`
- [x] Check all `// BUG: ` comments in the fuzzer again to ensure that nothing was missed / is unnecessary or out of place


juntyr:
@torkleyy Holy dang, serde attributes are a mess. This nightmare has really only unearthed two or three bugs in ron (around tagged enum tag identifier deserialising), many cases where we cannot roundtrip because serde's Content loses information (see https://github.com/serde-rs/serde/issues/1183), but MANY nightmares around serde attributes just plain not roundtripping themselves. I will try to add some tests and update our restrictions README in the PR, but I won't spend any time prettifying our fuzzer at this point. It will hopefully get nicer over time as the above restrictions are lifted.

Hopefully I can get this PR merged early next week and then finally, with the confidence that we now know about the limitations of serde attributes in ron, release v0.9.

src/parse.rs
```diff
@@ -943,22 +961,24 @@ impl<'a> Parser<'a> {
     }
 
     pub fn skip_ws(&mut self) -> Result<()> {
-        if self.src().is_empty() {
-            return Ok(());
-        }
-
-        if (self.cursor.pre_ws_cursor + self.cursor.last_ws_len) < self.cursor.cursor {
+        if (self.cursor.last_ws_len == WS_CURSOR_UNCLOSED_LINE)
+            || ((self.cursor.pre_ws_cursor + self.cursor.last_ws_len) < self.cursor.cursor)
```
juntyr, line 965:
```suggestion
        if (self.cursor.last_ws_len != WS_CURSOR_UNCLOSED_LINE)
            && ((self.cursor.pre_ws_cursor + self.cursor.last_ws_len) < self.cursor.cursor)
```

juntyr: Add a test for this

* calling after the end of a token with no trailing white space should still reset the token
* calling after trailing white space should not change anything
* calling after an unclosed comment shouldn't change anything either since the document is by definition already closed and the start of the white space was at / before the start of that comment

juntyr:
@torkleyy I think this is finally ready* (I want to merge and rebase on #512 first). Gosh this PR is huge ... and soooo cursed. And probably once the fuzzer changes lands I get many new edge case reports soon. But it is a small step *phew*

juntyr (commented):
LGTM

PR #512, opened 2023-10-02
Fixes #511: Allow deserialize_string for map fields originating from struct-formatted maps

### LINKED ISSUE #511: Problem deserializing maps when deserialization target expect owned String keys
I just encountered this somewhat strange behavior. Trying to deserialize a very simple RON string into a `serde_json::Value` fails with an `ExpectedIdentifier` error:

```rust
    let json: serde_json::Value = ron::from_str("(f1: 0, f2: 1)").unwrap();
```
I know that RON is not meant to be a self describing format and supports more types than JSON, but I expected this simple structure to be deserialized correctly; the identifiers should support deserializing to string keys.

The trigger of the error is that the `serde_json::Value` type expect owned `String` keys. After some investigating, it seems that the problem is related to the `id::Deserializer` variant, which supports `deserialize_str(..)`, but NOT `deserialize_string(..)`, which I find somewhat inconsistent. These methods should give the same result; the reason for having two methods is to give hints about expected string ownership (for performance reasons).

The following patch in `de/id.rs` file makes the code example work as expected:

```rust
// de::id.rs, line 147:

    fn deserialize_string<V>(self, visitor: V) -> Result<V::Value>
    where
        V: Visitor<'b>,
    {
        // Err(Error::ExpectedIdentifier)
        self.deserialize_identifier(visitor)  // Same as deserialize_str behavior
    }
```

grindvoll:
This commit fixes an incoherence problem in the Deserializer implementation for fields in struct-encoded maps. `deserialize_str` was accepted, while `deserialize_string` was rejected. This behavior caused a problem when deserializing e.g. `serde_json::Values` from RON-files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix serialising reserved identifiers `true`, `false`, `Some`, `None`, `inf`[`f32`|`f64`], and `Nan`[`f32`|`f64`] ([#487](https://github.com/ron-rs/ron/pull/487))
 - Disallow unclosed line comments at the end of `ron::value::RawValue` ([#489](https://github.com/ron-rs/ron/pull/489))
 - Fix parsing of struct/variant names starting in `None`, `Some`, `true`, or `false` ([#499](https://github.com/ron-rs/ron/pull/499))
+- Fix deserialising owned string field names in struct formatted maps, allowing deserialization of `serde_json::Value` values ([#511](https://github.com/ron-rs/ron/pull/512))
 
 ### Miscellaneous
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix serialising reserved identifiers `true`, `false`, `Some`, `None`, `inf`[`f32`|`f64`], and `Nan`[`f32`|`f64`] ([#487](https://github.com/ron-rs/ron/pull/487))
 - Disallow unclosed line comments at the end of `ron::value::RawValue` ([#489](https://github.com/ron-rs/ron/pull/489))
 - Fix parsing of struct/variant names starting in `None`, `Some`, `true`, or `false` ([#499](https://github.com/ron-rs/ron/pull/499))
-- Fix deserialising owned string field names in struct formatted maps, allowing deserialization of `serde_json::Value` values ([#511](https://github.com/ron-rs/ron/pull/512))
+- Fix deserialising owned string field names in structs, allowing deserializing into `serde_json::Value`s ([#511](https://github.com/ron-rs/ron/pull/512))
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix serialising reserved identifiers `true`, `false`, `Some`, `None`, `inf`[`f32`|`f64`], and `Nan`[`f32`|`f64`] ([#487](https://github.com/ron-rs/ron/pull/487))
 - Disallow unclosed line comments at the end of `ron::value::RawValue` ([#489](https://github.com/ron-rs/ron/pull/489))
 - Fix parsing of struct/variant names starting in `None`, `Some`, `true`, or `false` ([#499](https://github.com/ron-rs/ron/pull/499))
+- Fix deserialising owned string field names in struct formatted maps, allowing deserialization of `serde_json::Value` values ([#511](https://github.com/ron-rs/ron/pull/512))
```
juntyr, line 43:
```suggestion
- Fix deserialising owned string field names in structs, allowing deserializing into `serde_json::Value`s ([#511](https://github.com/ron-rs/ron/pull/512))
```

juntyr:
@grindvoll Thanks for the PR! It looks good to me once codecov is 100% and with the small CHANGELOG tweak.

tests/511_deserialize_any_map_string_key.rs
```diff
@@ -0,0 +1,51 @@
...
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            struct CVisitor;
+            impl<'de> serde::de::Visitor<'de> for CVisitor {
+                type Value = CustomMap;
+                fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                    write!(formatter, "a map with string keys and values")
+                }
```
juntyr, line 20:
```suggestion
                type Value = CustomMap;
                
                // GRCOV_EXCL_START
                fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                    write!(formatter, "a map with string keys and values")
                }
                // GRCOV_EXCL_STOP
```

juntyr: This small change should be sufficient to get coverage back to 100%.

juntyr:
@grindvoll Could you add a second test function to the new file, which just checks your motivating case, e.g.

```rust
#[test]
fn ron_struct_as_json_map() {
    let json: serde_json::Value = ron::from_str("(f1: 0, f2: 1)").unwrap();
    assert_eq!(
        json,
        serde_json::Value::Object(
            [
                (
                    String::from("f1"),
                    serde_json::Value::Number(serde_json::Number::from(0))
                ),
                (
                    String::from("f2"),
                    serde_json::Value::Number(serde_json::Number::from(1))
                ),
            ]
            .into_iter()
            .collect()
        )
    );
}
```

PR #522, opened 2024-01-03
Add `explicit_struct_names` Extension

### LINKED ISSUE #519: Option to Expect Specific Struct Name Upon Deserialization
I think it would be helpful to be able to enforce that a specific struct name is present when deserializing a struct. I mainly want to enforce this as part of style in my game's assets.

SylvKT:

Fixes #519


src/de/mod.rs
```diff
@@ -723,6 +723,10 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
 
         self.parser.skip_ws()?;
 
+        if self.parser.exts.contains(Extensions::EXPLICIT_STRUCT_NAMES) && name.len() == 0 {
```
juntyr, line 726: With this, you are checking if the serde-given struct name is empty or not, but not whether the user has provided it or not. In line 721 just above, when we call `self.parser.consume_struct_name(name)?;`, the struct name is consumed (if it is there we eat it, if it's not we do nothing). I think that the check should rather be in this function such that continuing when the struct name isn't given is no longer allowed with the extension.

juntyr (changes requested):

src/de/mod.rs
SylvKT: I realized this while reading the codebase before I saw this, but thanks. I've implemented this logic in `Parser::consume_struct_name` and created a test for it.

src/extensions.rs
```diff
@@ -40,19 +43,25 @@ mod tests {
...
+        roundtrip_extensions(Extensions::UNWRAP_NEWTYPES | Extensions::EXPLICIT_STRUCT_NAMES);
         roundtrip_extensions(Extensions::IMPLICIT_SOME | Extensions::UNWRAP_VARIANT_NEWTYPES);
+        roundtrip_extensions(Extensions::IMPLICIT_SOME | Extensions::EXPLICIT_STRUCT_NAMES);
+        roundtrip_extensions(Extensions::UNWRAP_VARIANT_NEWTYPES | Extensions::EXPLICIT_STRUCT_NAMES);
         roundtrip_extensions(
             Extensions::UNWRAP_NEWTYPES
                 | Extensions::IMPLICIT_SOME
-                | Extensions::UNWRAP_VARIANT_NEWTYPES,
+                | Extensions::UNWRAP_VARIANT_NEWTYPES
+                | Extensions::EXPLICIT_STRUCT_NAMES,
```
SylvKT, line 64: I'm not sure if i got this right.

src/parse.rs
```diff
@@ -666,7 +666,11 @@ impl<'a> Parser<'a> {
 
     pub fn consume_struct_name(&mut self, ident: &'static str) -> Result<bool> {
         if self.check_ident("") {
-            return Ok(false);
+            if self.exts.contains(Extensions::EXPLICIT_STRUCT_NAMES) {
+                return Err(Error::ExpectedStructName(ident.to_string()))
```
SylvKT, line 670: Here's where I implemented the extension.

tests/522_explicit_struct_names.rs
```diff
@@ -0,0 +1,43 @@
...
+        id: (3),
+    )"#;
+    let foo = options.from_str::<Foo>(content);
+    match foo.expect_err(EXPECT_ERROR_MESSAGE).code {
+        Error::ExpectedStructName(_) => {},
+        err => panic!("{INCORRECT_ERROR_MESSAGE} \"{err}\""),
+    }
+
+    // phase 3 (use content from phase 2)
+    let _foo = from_str::<Foo>(content).unwrap();
```
SylvKT, line 42: I use three phases: one to check if the extension works with regular `struct`s, one to check if it works with `newtype structs`, and one to ensure no errors occur with the default extensions. Should I add one to check if it works with `tuple structs` as well?

SylvKT (commented):
added some questions and comments

SylvKT:
Before you merge, I'm adding some documentation

src/extensions.rs
juntyr: What we really want to test here is the powerset of all extensions (while that is still possible). Perhaps there is a way to use the bitset API to iterate over the powerset and then invoke the test function for each subset?

src/error.rs
```diff
@@ -281,6 +282,7 @@ impl fmt::Display for Error {
                 "Exceeded recursion limit, try increasing `ron::Options::recursion_limit` \
                 and using `serde_stacker` to protect against a stack overflow",
             ),
+            Error::ExpectedStructName(ref name) => write!(f, "Expected a struct name `{}`, but none was found", name),
```
juntyr, line 285:
```suggestion
            Error::ExpectedStructName(ref name) => write!(f, "Expected the explicit struct name `{}`, but none was found", name),
```

tests/522_explicit_struct_names.rs
juntyr: Sure, that's a good idea!

juntyr (commented):
Overall, the changes all look good - thank you for also adding documentation!

One question that we should still resolve is how this extension relates to the `PrettyConfig::struct_names` option. I think it would make sense that enabling the extension during serialisation forces all struct names to be emitted (even when we're not in pretty serialising mode). Do we then still need the `PrettyConfig::struct_names` option? I think if we can still emit struct names with and without (as two distinct options) emitting the explicit `#[enable(explicit_struct_names)]` marker, we can probably also axe the config option.

SylvKT:
I think it would be more intuitive to keep the serialization configuration options separate from the deserialization extension. If `PrettyConfig` is only used for serialization, then it would make sense to put all of the serialization options in one place. If we go that route, we should add a "see also" link to the documentation for `PrettyConfig::struct_names` and vice versa.

src/extensions.rs
SylvKT: I've figured out a clean way to do this and implemented it.

SylvKT:
> If `PrettyConfig` is only used for serialization, then it would make sense to put all of the serialization options in one place. If we go that route, we should add a "see also" link to the documentation for `PrettyConfig::struct_names` and vice versa.

It turns out `PrettyConfig` does use `Extensions`, so I think we should just drop `PrettyConfig::struct_names`.

juntyr:
> I think it would be more intuitive to keep the serialization configuration options separate from the deserialization extension. If `PrettyConfig` is only used for serialization, then it would make sense to put all of the serialization options in one place. If we go that route, we should add a "see also" link to the documentation for `PrettyConfig::struct_names` and vice versa.

So far, all extensions have served a dual purpose of specifying a variation of the RON format and enabling it during both serialisation and deserialisation, so that you can produce a RON document with the same options (that’s how the Options struct originated) that you will apply during parsing. Therefore, I think it would be good for consistency to keep that dual functionality.

So far, explicit struct names have just been a prettifying option, but this PR obviously changes that. If we keep the PrettyConfig option as well, the extension would be used in non-pretty mode, but we’d need to decide how to merge the extension and the pretty config - or-ing them makes most sense.


Either way, whether we keep the PrettyConfig option or not, I agree with you that we should add a short explanation to each item.

src/extensions.rs
```diff
@@ -42,17 +45,10 @@ mod tests {
...
-        roundtrip_extensions(Extensions::UNWRAP_NEWTYPES | Extensions::IMPLICIT_SOME);
-        roundtrip_extensions(Extensions::UNWRAP_NEWTYPES | Extensions::UNWRAP_VARIANT_NEWTYPES);
-        roundtrip_extensions(Extensions::IMPLICIT_SOME | Extensions::UNWRAP_VARIANT_NEWTYPES);
-        roundtrip_extensions(
-            Extensions::UNWRAP_NEWTYPES
-                | Extensions::IMPLICIT_SOME
-                | Extensions::UNWRAP_VARIANT_NEWTYPES,
-        );
+        // iterate over the powerset of all extensions (i.e. every possible combination of extensions)
+        for bits in Extensions::empty().bits()..=Extensions::all().bits() {
```
juntyr, line 49: That’s a very clever way to do it - thank you for figuring it out!

juntyr:
Could you also bring back the coverage to 100%? There is a test in the error file where you can just add another case. Then, in your own test, you could either try to use asserts or manually disable coverage for the unreachable panics (search for GRCOV_EXCL_ in other tests for examples).

All in all, thank you so much for working on this addition! Once these last minor nits are addressed, it’ll be ready to be merged.

SylvKT:
I believe I've fixed the formatting and coverage now.

src/error.rs
```diff
@@ -281,6 +282,11 @@ impl fmt::Display for Error {
                 "Exceeded recursion limit, try increasing `ron::Options::recursion_limit` \
                 and using `serde_stacker` to protect against a stack overflow",
             ),
+            Error::ExpectedStructName(ref name) => write!(
+                f,
+                "Expected the explicit struct name `{}`, but none was found",
```
juntyr, line 287:
```suggestion
                "Expected the explicit struct name {}, but none was found",
```

src/error.rs
```diff
@@ -281,6 +282,11 @@ impl fmt::Display for Error {
...
+                name
```
juntyr, line 288:
```suggestion
                Identifier(name)
```

src/ser/mod.rs
```diff
@@ -163,10 +163,13 @@ impl PrettyConfig {
 
     /// Configures whether to emit struct names.
     ///
+    /// Note that when using the `explicit_struct_names` extension, this method will use an OR operation on the extension and the [`PrettyConfig`] option. See also [`Extensions::EXPLICIT_STRUCT_NAMES`] for the extension equivalent.
+    ///
     /// Default: `false`
     #[must_use]
     pub fn struct_names(mut self, struct_names: bool) -> Self {
-        self.struct_names = struct_names;
+        self.struct_names =
```
juntyr, line 171: I think this is the wrong place to do the OR-ing - what about the struct_names helper function in the Serializer (see https://github.com/ron-rs/ron/blob/00a65cdd9e01cfc54066387069f5218b09a74378/src/ser/mod.rs#L642) which should now look at the pretty config, pretty config extensions, and default extensions (is there already a union of those somewhere?)

tests/522_explicit_struct_names.rs
```diff
@@ -0,0 +1,62 @@
...
+    // phase 3 (tuple structs)
+    let content_tuple = r#"Foo(
+        id: Id(3),
+        position: (0.0, 8.72),
+    )"#;
+    let foo = options.from_str::<Foo>(content_tuple);
+    assert_eq!(
+        foo.expect_err(EXPECT_ERROR_MESSAGE).code,
+        Error::ExpectedStructName("Position".to_string())
+    );
```
juntyr, line 56: Could you also add a test for this extension combined with unwrap variant newtypes, e.g. that Enum::Variant(Struct { a: 42 }) is still serialised to and deserialised from Variant(a: 42) ? Sorry for only just thinking of this now.

juntyr:
Ok, just a couple more minor nits - thank you again for implementing the extension!

src/ser/mod.rs
```diff
@@ -636,10 +638,15 @@ impl<W: fmt::Write> Serializer<W> {
         Ok(())
     }
 
+    /// Checks if struct names should be emitted
+    ///
+    /// Note that when using the `explicit_struct_names` extension, this method will use an OR operation on the extension and the [`PrettyConfig::struct_names`] option. See also [`Extensions::EXPLICIT_STRUCT_NAMES`] for the extension equivalent.
     fn struct_names(&self) -> bool {
         self.pretty
```
juntyr, line 645: I think this should rather be:
```rust
self.extensions().contains(Extensions::EXPLICIT_STRUCT_NAMES) || self.pretty.as_ref().map_or(false, |(pc, _)| pc.struct_names)
```

src/ser/mod.rs
juntyr, line 1: I just also noticed https://github.com/ron-rs/ron/blob/08ebb5d850177053f09854d6c6769561c44e3078/src/ser/mod.rs#L418-L431 again. The new extension should be added here as well - perhaps this code could now be refactored to iterate over all enabled non-default extensions and print them by getting their lower-case-name (so far this is just specified separately by the parse method and this printing code, but perhaps the Extensions bitset could combine both and also have a function to get this name for any given variant?).

juntyr:
> I believe I've fixed the formatting and coverage now.

Yes, thank you! It seems one unrelated #[derive(Debug)] line has regressed in https://app.codecov.io/gh/ron-rs/ron/pull/522/blob/tests/non_string_tag.rs though

src/extensions.rs
```diff
@@ -7,28 +7,35 @@ bitflags::bitflags! {
...
+        ///
+        /// During deserialization, this extension requires that structs' names are stated explicitly.
+        const EXPLICIT_STRUCT_NAMES = 0x8;
     }
 }
 // GRCOV_EXCL_STOP
 
 impl Extensions {
-    /// Creates an extension flag from an ident.
+    /// Creates an extension flag from an idents.
```
juntyr, line 19:
```suggestion
    /// Creates an extension flag from an ident.
```

juntyr (approved):
LGTM!

tests/522_explicit_struct_names.rs
SylvKT: consolidated with phase 5 (serialization/deserialization equivalence)

juntyr:
@SylvKT Thank you so much again for implementing this extension!

PR #541, opened 2024-06-11
Use/fix ron code blocks

spenserblack:
* [ ] I've included my change in `CHANGELOG.md` (I don't think this counts as a "notable change")

Just a heads-up that GitHub now supports RON. This PR
- changes the `rust,ignore` code blocks to `ron`
- changes the ` ron` (leading space) code blocks to `ron`

The downside of this PR is that, while GitHub now detects RON, other tools that render the markdown might not (like crates.io). So perhaps for that reason the `README.md` should be left alone and only the leading spaces should be fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ structs, enums, tuples, arrays, generic maps, and primitive values.
 
 ## Example
 
-```rust,ignore
+```ron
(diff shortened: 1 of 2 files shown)


juntyr (approved):
LGTM

PR #544, opened 2024-08-05
Add provisional metadata ser in named field pos

voidentente:
## Motive

It would be useful to be able to serialize additional metadata, i.e. data that is not scoped to the data representation itself (specifically, comments/documentation regarding the implications of a field/its possible values/valid range/etc).

Combined with crates like [documented](https://crates.io/crates/documented), this would (for example) allow for auto-documented configuration files.

This PR aims to provide a very lightweight (provisional) support for serializing metadata in named field position.

## Serialization

Metadata should be provided to the serializer through `PrettyConfig`, since it's targeted at human readability.

The only relevant serialization call is `<Compound as ser::SerializeStruct>::serialize_struct`.

### Why not in type position?

Distinguishing field position metadata from type position metadata is awkward. It's possible, but awkward:

```ron
/// type position
(
    /// field position
    ident: 
    /// type position??
    (
        /// field position
        a: (),
        b: (),
    )
)
```

### Why not in unit structs?

Unit structs are values:

```ron
(
    /// field position
    unit: 
    /// type position?
    (
        // no fields, since it's unit..
    ),
)
```

Without type position, since there's no fields, there's nothing to do.

### Why not in tuple structs?

Tuple structs have unnamed fields, which would make it appear as if values had metadata:

```ron
(
    /// field position
    tuple: (
        /// field position..?
        0,
        "",
    )
)
```

### Why not in newtype structs?

Same as with tuple structs.

### Why not in variants?

Variants are values, and suffer from the same problem as type position:

```ron
(
    /// field position
    variant: 
    /// type position?
    /// variant position??
    Some(13)
)
```

### Deserialization

Metadata is skipped during deserialization. This is given by using the comment syntax.

### Open Questions

- Where should the metadata live during serialization? In `Serializer`? `PrettyConfig`? Should it be skipped when serializing `PrettyConfig`?

- What to use to mark metadata? To keep backwards compatibility, either `//` or `///` are suitable, with `///` additionally offering forwards compatibility to distinguish metadata from plain comments.

- How to avoid overlaps? If two structs have a field that share a name, there’s no way to differentiate them. Additional data would have to be tracked to distinguish these namespaces. Because this might be surprising, I'd say this should stay a draft until this has a solution.

    Example of an overlap:

    ```rs
    mod inner {
        pub struct Definition {
            a: usize,
        }
    }

    struct Definition {
        a: usize,
        inner: inner::Definition,
    }
    ```

### Example Usage

```rs
use ron::ser::PrettyConfig;
use serde_derive::{Deserialize, Serialize};

fn main() {
    #[derive(Serialize, Deserialize)]
    pub enum C {
        Variant,
    }

    #[derive(Serialize, Deserialize)]
    pub struct B {
        pub c: C,
    }

    #[derive(Serialize, Deserialize)]
    struct A {
        a: usize,
        b: B,
    }

    let value = A {
        a: 0,
        b: B { c: C::Variant },
    };

    let mut config = PrettyConfig::default();

    config.meta.insert("a", "this is paired with a");
    config.meta.insert("b", "this is paired with b");
    config.meta.insert("c", "this is paired with c");

    let s = ron::ser::to_string_pretty(&value, config).unwrap();

    println!("{s}");

    assert!(ron::de::from_str::<A>(&s).is_ok());
}
```
```
(
    /// this is paired with a
    a: 0,
    /// this is paired with b
    b: (
        /// this is paired with c
        c: Variant,
    ),
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Breaking: Enforce that ron always writes valid UTF-8 ([#488](https://github.com/ron-rs/ron/pull/488))
 - Add convenient `Value::from` impls ([#498](https://github.com/ron-rs/ron/pull/498))
 - Add new extension `explicit_struct_names` which requires that struct names are included during deserialization ([#522](https://github.com/ron-rs/ron/pull/522))
+- Add new metadata serialization support in named field position via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
 
 ### Format Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Breaking: Enforce that ron always writes valid UTF-8 ([#488](https://github.com/ron-rs/ron/pull/488))
 - Add convenient `Value::from` impls ([#498](https://github.com/ron-rs/ron/pull/498))
 - Add new extension `explicit_struct_names` which requires that struct names are included during deserialization ([#522](https://github.com/ron-rs/ron/pull/522))
-- Add new metadata serialization support in named field position via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
+- Add new path-based field metadata serialization support via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
 - Breaking: Change `PrettyConfig` so that `new_line`, `indentor` and `separator` are all `Cow<'static, str>` instead of `String` ([#546](https://github.com/ron-rs/ron/pull/546))
 
 ### Format Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add convenient `Value::from` impls ([#498](https://github.com/ron-rs/ron/pull/498))
 - Add new extension `explicit_struct_names` which requires that struct names are included during deserialization ([#522](https://github.com/ron-rs/ron/pull/522))
 - Add new metadata serialization support in named field position via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
+- Breaking: Change `PrettyConfig` so that `new_line`, `indentor` and `separator` are all `Cow<'static, str>` instead of `String` ([#546](https://github.com/ron-rs/ron/pull/546))
 
 ### Format Changes
 
--- a/src/meta.rs
+++ b/src/meta.rs
@@ -1,0 +1,22 @@
+use std::collections::HashMap;
+
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+pub struct Meta {
+    inner: HashMap<String, String>,
+}
+
+impl Meta {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn insert(&mut self, k: impl Into<String>, v: impl Into<String>) {
+        self.inner.insert(k.into(), v.into());
+    }
+
+    pub fn get(&self, k: &str) -> Option<&str> {
+        self.inner.get(k).map(String::as_str)
+    }
+}
--- a/src/meta.rs
+++ b/src/meta.rs
@@ -2,28 +2,115 @@ use std::collections::HashMap;
 
 use serde_derive::{Deserialize, Serialize};
 
+/// Path-based metadata to serialize with a value.
+///
+/// Path-based in this context means that the metadata is linked
+/// to the data in a relative and hierarchical fashion by tracking
+/// the current absolute path of the field being serialized.
+///
+/// # Example
+///
+/// ```
+/// # use ron::{ser::PrettyConfig, meta::Field};
+///
+/// #[derive(serde::Serialize)]
+/// struct Creature {
+///     seconds_since_existing: usize,
+///     linked: Option<Box<Self>>,
+/// }
+///
+/// let mut config = PrettyConfig::default();
+///
+/// config
+///     .meta
+///     .field
+///     // The path meta defaults to no root structure,
+///     // so we either provide a prebuilt one or initialize
+///     // an empty one to build.
+///     .get_or_insert_with(Field::empty)
+///     .build_fields(|fields| {
+///         fields
+///             // Get or insert the named field
+///             .field("seconds_since_existing")
+///             .with_doc("Outer seconds_since_existing");
+///         fields
+///             .field("linked")
+///             // Doc metadata is serialized preceded by three forward slashes and a space for each line
+///             .with_doc("Optional.\nProvide another creature to be wrapped.")
+///             // Even though it's another Creature, the fields have different paths, so they are addressed separately.
+///             .build_fields(|fields| {
+///                 fields
+///                     .field("seconds_since_existing")
+///                     .with_doc("Inner seconds_since_existing");
+///             });
+///     });
+///
+/// let value = Creature {
+///     seconds_since_existing: 0,
+///     linked: Some(Box::new(Creature {
+///         seconds_since_existing: 0,
+///         linked: None,
+///     })),
+/// };
+///
+/// let s = ron::ser::to_string_pretty(&value, config).unwrap();
+///
+/// assert_eq!(s, r#"(
+///     /// Outer seconds_since_existing
+///     seconds_since_existing: 0,
+///     /// Optional.
+///     /// Provide another creature to be wrapped.
+///     linked: Some((
+///         /// Inner seconds_since_existing
+///         seconds_since_existing: 0,
+///         linked: None,
+///     )),
+/// )"#);
+/// ```
+///
+/// # Identical paths
+///
+/// Especially in enums and tuples it's possible for fields
+/// to share a path, thus being unable to be addressed separately.
+///
+/// ```no_run
+/// enum Kind {
+///     A {
+///         field: (),
+///     },  // ^
+///         // cannot be addressed separately because they have the same path
+///     B { // v
+///         field: (),
+///     },
+/// }
+/// ```
+///
+/// ```no_run
+/// struct A {
+///     field: (),
+/// }
+///
+/// struct B {
+///     field: (),
+/// }
+///
+/// type Value = (
+///     A,
+///  // ^
+///  // These are different types, but they share two fields with the same path: `buf` and `len`
+///  // v
+///     B,
+/// );
+/// ```
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
-pub struct Meta {
-    fields: Fields,
+pub struct PathMeta {
+    pub field: Option<Field>,
 }
 
-impl Meta {
-    /// Get a reference to the named field position metadata.
-    #[must_use]
-    pub fn fields(&self) -> &Fields {
-        &self.fields
-    }
-
-    /// Get a mutable reference to the named field position metadata.
-    pub fn fields_mut(&mut self) -> &mut Fields {
-        &mut self.fields
-    }
-}
-
-/// The metadata and inner [Fields] of a field.
+/// The metadata and inner [`Fields`] of a field.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
 pub struct Field {
-    meta: String,
+    doc: String,
     fields: Option<Fields>,
 }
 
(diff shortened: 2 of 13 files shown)


voidentente:
Regarding the collision problem: Binding the meta to the struct name as well and not just field name significantly reduces the risk of collision, but does not eliminate it. `A { a }` would still be indistinguishable from `inner::A { a }`.

Since the type isn't accessible from serialization however, the only way to fully eliminate the collision problem is by not relying on type binding in the first place. One alternative might be to supply the meta as a structured hierarchy.

This requires the serializer to keep track of which field(s) are currently entered. 
My API currently looks something like this:

```rs
fn main() {
    #[derive(Serialize)]
    struct A {
        a: usize,
        b: inner::A,
        c: usize,
    }

    mod inner {
        use serde_derive::Serialize;
        #[derive(Serialize)]
        pub struct A {
            pub a: usize,
        }
    }

    let mut config = PrettyConfig::default();

    {
        let meta = &mut config.meta;

        {
            let field = meta.field_mut_or_default("a");
            field.set_meta("field a");
        }

        {
            let field = meta.field_mut_or_default("b");
            field.set_meta("field b");

            field.set_inner({
                let mut fields = Fields::new();
                fields.field_mut_or_default("a").set_meta("inner field a");
                fields
            });
        }

        {
            let field = meta.field_mut_or_default("c");
            field.set_meta("field c");
        }
    }

    let value = A {
        a: 0,
        b: inner::A { a: 0 },
        c: 3,
    };

    let s = ron::ser::to_string_pretty(&value, config).unwrap();

    println!("{s}");
}
```
```
(
    /// field a
    a: 0,
    /// field b
    b: (
        /// inner field a
        a: 0,
    ),
    /// field c
    c: 3,
)
```

voidentente:
Alright, I think that should clear up the collision problem. Support for more positions can be added later seamlessly. Going with `PrettyConfig` and `///` seems like a reasonable choice to me. I'll un-draft this because the impact of this PR on the crate should be rather minimal

juntyr:
Thank you @voidentente for your PR and the extensive motivation!

I absolutely agree with the motivation for adding attributes in general and doc comments as a first step towards them.

What I'm unsure about is the focus on type-based docs instead of value-based docs. As you very thoroughly lay out, there are almost no uniquely type-based places for docs to appear, with struct-like fields being the exception, most other places feel more like they would be used for value-docs.

Furthermore, the API for type-based docs feels a bit brittle as there is a possibility for name clashes or the structure of the type names needs to be separately encoded again.

What I would prefer is a value-based docs (later generalised to attributes) API that is more general and can be utilised to then build a type-based API on top of it (but probably not inside RON but another crate). In particular, I would suggest that stylistically, doc comments generally go on the line above the value and we add a special case for fields so that they are serialised as follows:

```ron
(
    /// my value comment
    a: 42,
)
```

What I think might work well is an API similar to `serde_spanned` (which I've wanted to support in RON for a while but haven't gotten to), that is:
- a new separate `serde_meta` or `serde_value_attrs` crate (I don't think one exists already) that publicly exports a type `pub struct Meta<'a, T> { meta: Cow<'a, [(Cow<'a, str>, Cow<'a, str>)]>, value: T }` and hidden-exports some helper functions to identify when this type is being serialised and deserialised (taking `serde_spanned` for heavy inspiration)
- add serialising support such that `("doc", "my-doc")` attributes are serialised as `/// my-doc` (and error for any other attribute keys right now)
- add deserialising support such that doc comments are parsed, usually ignored, but provided when a struct that matches `Meta` is deserialised

Types that would always like to serialise the same doc comments (type-based) could then update their serialisation code to always serialise the same comment. Your proposed API of taking an existing data structure and serialising it with separate type comments could be built by wrapping the serialiser and injecting metadata structs whenever a matching struct is serialised.

What are your thoughts?

voidentente:
Well, I expressly wanted to keep the extent of this PR as scoped as possible. It's based entirely on top of the comment syntax, because I'm unsure of the idiomacy of deserialization of metadata. If metadata is capable of deserialization, it wouldn't be just for human readability. At that point, the difference between metadata and normal data becomes muddy.

You bring up other attribute kinds, which would require a new syntax and thus support by the deserializer. I'm hesitant about this, because it would be a breaking change (unless some piggybacking happens, like `//kind/ value` or similar).

The primary motivation (at least for me here) is serializing documentation. There's no need to be capable of deserializing metadata in order to update it; the user should be able to dictate what metadata to put where regardless of the previous state of the document.

Since this PR should be entirely non-breaking, it'd be a first good step to adding support. It wouldn't be forwards-compatible with an extended format, but that's fine if this part of the API is marked as unstable. I'm not knowledgeable enough about the serde ecosystem to transform this into a fully-featured, integrated, stable API from the get-go.

> What I'm unsure about is the focus on type-based docs instead of value-based docs. As you very thoroughly lay out, there are almost no uniquely type-based places for docs to appear, with struct-like fields being the exception, most other places feel more like they would be used for value-docs.
> 
> Furthermore, the API for type-based docs feels a bit brittle as there is a possibility for name clashes or the structure of the type names needs to be separately encoded again.

Type position is usually a subset of field position, which is why I chose to neglect it. Serializing in a way that differentiates it between field position metadata would hurt readability.

```ron
/// type position
Type (
    /// field position
    a: 
    /// type position
    0,
    /// field position
    opt: 
    /// type position
    Some(13),
    /// field position
    inner: 
    /// type position
    Other {
        /// field position
        value: 
        /// type position
        (),
    }
)
```

The collision problem should be entirely resolved by using hierarchy. The serializer internally keeps track of which field is entered, allowing to differentiate fields with the same name based on context.

The (field position) metadata for the above RON is currently expressed like this:

```
["a"]              = "field position"
["opt"]            = "field position"
["inner"]          = ..
["inner"]["value"] = ..
```

juntyr:
> Well, I expressly wanted to keep the extent of this PR as scoped as possible. It's based entirely on top of the comment syntax, because I'm unsure of the idiomacy of deserialization of metadata. If metadata is capable of deserialization, it wouldn't be just for human readability. At that point, the difference between metadata and normal data becomes muddy.

I appreciate the minimal initial approach and the focus on doc comments as a first step! And it’s true that treating metadata as data in the serde model might not be suited for all use cases, where your proposed more ad-hoc API would work very well.

Your proposed API is thus growing on me and I think it could be supported long-term, even if it’s internal implementation might at some point switch to a proper attribute system.

> You bring up other attribute kinds, which would require a new syntax and thus support by the deserializer. I'm hesitant about this, because it would be a breaking change (unless some piggybacking happens, like `//kind/ value` or similar).

Ron already has attributes, though they’re currently only used to enable features and are only allowed at the top of the document. This could simply be expanded, and just like in Rust, doc comments would be supported as /// and #[doc = “”].

juntyr:
What this PR would still need is tests to ensure full coverage and to test how field docs can be added e.g. to a struct nested in an enum inside a vec

voidentente:
> how field docs can be added e.g. to a struct nested in an enum inside a vec

TL;DR: The hierarchy is not absolute, but relative to entered structs with named fields. The test file asserts that addressing fields of a struct nested in an enum inside a vec is structurally equivalent to addressing a struct.

juntyr:
API-wise, I think the meta field on PrettyConfig should be of an new Meta type, which then exposes a fields() method, so that we could add support for different accessors later.

Could there be a test for a tuple (Person, Pet) and how the name clashes are handled there?

voidentente:
@juntyr Any more thoughts on this?

src/ser/mod.rs
```diff
@@ -1329,6 +1337,30 @@ impl<'a, W: fmt::Write> ser::SerializeStruct for Compound<'a, W> {
 
         if !self.ser.compact_structs() {
             self.ser.indent()?;
+
+            if let Some((ref config, _)) = self.ser.pretty {
+                let mut iter = self.ser.field_memory.iter();
+
+                let init = iter.next().and_then(|name| config.meta.fields().get(name));
+                let field = iter
```
juntyr, line 1345: Would it be possible to not have this loop for every compound? What I'm imagining is a stack of borrows into the metadata instead of a stack of field names, so that when we serialise a field, we already know if it has metadata and don't have to search any further

src/meta.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub fn new(meta: impl Into<String>, fields: Option<Fields>) -> Self {
+        Self {
+            meta: meta.into(),
+            fields,
+        }
+    }
+
+    /// Get the metadata of this field.
+    #[must_use]
+    pub fn meta(&self) -> &str {
```
juntyr, line 50: Since I'm still thinking about generalising this feature in the future, I think that meta should not be string-typed but that there should be a meta struct with private members (for now just a string) and a method to set a doc comment (since that's the metadata we support right now). A later extension could then add further private fields and public methods to this struct

src/meta.rs
```diff
@@ -0,0 +1,274 @@
+use std::collections::HashMap;
+
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+pub struct Meta {
```
juntyr, line 6: This metadata is path-based (it reminds me of json path selectors), maybe we can somehow include that in the name

juntyr (changes requested):

juntyr:
Sorry for the delay in the review, things have been hard

src/ser/mod.rs
voidentente: If we're talking about storing an `Option<&Meta>` in the Serializer instead of a stack of field names, it'd be self-referential, which the borrow checker will not particularly like.

src/meta.rs
voidentente: I'd be as simple as calling it `PathMeta`, is that okay?

src/ser/mod.rs
juntyr: It would be an Vec<&Fields>, but your point stands … is there another way we could avoid the quadratic runtime here?

src/meta.rs
juntyr: Yes :)

src/ser/mod.rs
voidentente: We could avoid borrowing by mutating the PrettyConfig in place during serialization.

In serialize_field:
Before serialization:
- Remove the field we are entering from the config.
- Swap the config's current field with the removed field.
During serialization:
- The field is directly accessed, no further processing.
After serialization:
- Swap the config's current field with the original field.
- Insert the removed field back into the config.

This moves the head of the config so it always points directly at the field we're currently serializing. Performance-wise, it'd just be swapping of `Field` (<100 bytes).

juntyr: That sounds like a good plan to me

voidentente:
The path traversal should now be constant time. I'm not super happy with the public API, but it's prone to change anyway.

src/ser/mod.rs
```diff
@@ -1329,6 +1342,29 @@ impl<'a, W: fmt::Write> ser::SerializeStruct for Compound<'a, W> {
...
+                            buf.push_str("/// ");
+                            buf.push_str(line);
+                            buf.push('\n');
+                            buf
+                        })
+                        .collect();
+
+                    for doc_line in doc_lines {
+                        self.ser.output.write_str(&doc_line)?;
+                        self.ser.indent()?;
```
voidentente, line 1364: Note that the `indent` method overcaptures and borrows the entire serializer mutably (it only needs to borrow the output mutably), resulting in an avoidable heap allocation when we write the metadata. Resolving this would probably involve either inlining `indent` or making it a freestanding function, but that's outside of the scope of this PR, so I left it alone.

voidentente:
I'll also throw in that `Fields` currently uses SipHash, which might be overkill. We could get a minor performance increase with AHash.

juntyr (changes requested):

juntyr:
Thank you @voidentente for your continuous work on this PR - just a few more nits and I'm happy to land this

src/meta.rs
```diff
@@ -0,0 +1,365 @@
...
+///     A,
+///  // ^
+///  // These are different types, but they share two fields with the same path: `buf` and `len`
+///  // v
+///     B,
+/// );
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Default)]
+pub struct PathMeta {
+    pub field: Option<Field>,
```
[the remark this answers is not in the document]
voidentente: To be honest, `PathMeta` and `Option<Field>` are currently synonymous, and I'm considering removing this layer of abstraction rather than encapsulate it further. The `path_meta` field in `PrettyConfig` should probably just be `Option<Field>`.

voidentente:
I moved the PR to the `ser` module just to make it more obvious that it only performs serialization, and renamed it and the test file to path_meta. `PrettyConfig` now contains it directly, which should be more intuitive.

src/ser/mod.rs
```diff
@@ -489,17 +503,6 @@ impl<W: fmt::Write> Serializer<W> {
         Ok(())
     }
 
-    fn indent(&mut self) -> fmt::Result {
```
juntyr, line 492: I think I'd still like to keep this method as a thin wrapper around the new function

src/ser/mod.rs
```diff
@@ -378,6 +383,15 @@ pub struct Serializer<W: fmt::Write> {
     implicit_some_depth: usize,
 }
 
+fn indent<W: fmt::Write>(mut output: W, config: &PrettyConfig, pretty: &Pretty) -> fmt::Result {
```
juntyr, line 386: Could this take a mutable reference to output instead, since we're anyways always passing one?

juntyr (approved):
I just have two minor nits, but everything else looks great :)

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Breaking: Enforce that ron always writes valid UTF-8 ([#488](https://github.com/ron-rs/ron/pull/488))
 - Add convenient `Value::from` impls ([#498](https://github.com/ron-rs/ron/pull/498))
 - Add new extension `explicit_struct_names` which requires that struct names are included during deserialization ([#522](https://github.com/ron-rs/ron/pull/522))
+- Add new metadata serialization support in named field position via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
```
juntyr, line 24:
```suggestion
- Add new path-based field metadata serialization support via `PrettyConfig` ([#544](https://github.com/ron-rs/ron/pull/544))
```

src/ser/mod.rs
```diff
@@ -109,6 +111,8 @@ pub struct PrettyConfig {
     pub compact_maps: bool,
     /// Enable explicit number type suffixes like `1u16`
     pub number_suffixes: bool,
+    /// Additional metadata to serialize
```
juntyr, line 114:
```suggestion
    /// Additional path-based field metadata to serialize
```

juntyr (approved):
LGTM

juntyr:
Thank you @voidentente for your work on this feature!

PR #546, opened 2024-09-11
Make use of `Cow<'static, str>` in `PrettyConfig`

### LINKED ISSUE #545: Please use `Cow<'static, str>` for configuration, rather than `String`.
**Please use `Cow<'static, str>` for configuration, rather than `String`.**

This way, we will not have to allocate new `String`s all the time even when the strings are always the same.
This will also make it possible to construct configurations in `const` contexts.

This will allow for cleaner code and slightly improved performance.

**How it is now:**

```rs
// This is a config that we will use repeatedly throughout the program.
// It must be inside the body of a function because allocating `String`s is not `const`.
let config = PrettyConfig::new()
    .indentor("\t".to_string())
    .struct_names(true);

// These all involve allocating new `String`s for the new line, indentor, and separator.
println!("{}", to_string_pretty(struct_1, config.clone()).unwrap());
println!("{}", to_string_pretty(struct_2, config.clone()).unwrap());
println!("{}", to_string_pretty(struct_3, config.clone()).unwrap());
```

**How I imagine it could be:**

```rs
// This can now be global if we want!
const CONFIG: PrettyConfig = PrettyConfig::new()
    .indentor(Cow::Borrowed("\t"))
    .struct_names(true);

// This is now cheaper and cleaner.
println!("{}", to_string_pretty(struct_1, CONFIG).unwrap());
println!("{}", to_string_pretty(struct_2, CONFIG).unwrap());
println!("{}", to_string_pretty(struct_3, CONFIG).unwrap());
```

*P.S. This is my first GitHub issue so sorry if it's, like, bad somehow.*

Coca162:
Fixes #545 by moving all `String` fields in `PrettyConfig` to `Cow<'static, str>`, builder methods are updated to take in `impl Into<Cow<'static, str>>`. I've also removed any conversions to strings in tests/other parts of the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Breaking: Enforce that ron always writes valid UTF-8 ([#488](https://github.com/ron-rs/ron/pull/488))
 - Add convenient `Value::from` impls ([#498](https://github.com/ron-rs/ron/pull/498))
 - Add new extension `explicit_struct_names` which requires that struct names are included during deserialization ([#522](https://github.com/ron-rs/ron/pull/522))
+- Breaking: Change `PrettyConfig` so that `new_line`, `indentor` and `separator` are all `Cow<'static, str>` instead of `String` ([#546](https://github.com/ron-rs/ron/pull/546))
 
 ### Format Changes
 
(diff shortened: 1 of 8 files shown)


juntyr:
Thank you @Coca162 for the PR and @maypop-the-dragon for bringing up the issue :)

Issue #550, opened 2024-10-29
Make `#[derive(Debug)]` output compatible with ron


HKalbasi:
This makes ron useful in deserializing output of programs using default Debug trait of Rust.

juntyr:
Do you mean making RON fully compatible with Rust syntax?

HKalbasi:
I mean making the output of `#[derive(Debug)]` a subset of ron syntax, so I can use `println!("{:?}", value)` in addition to normal serializing.

For example, `X { a: 1, b: 2 }` should be accepted in addition to the current `X ( a: 1, b: 2 )`. There is no need to drop support for the current form to make it Rust syntax compatible.

juntyr:
This *may* be possible to do. Ron has syntax for maps (since ron is a superset of JSON), so `{ a: 1, b: 2 }` would be parsed as a map from an enum with variants a and b to ints. I mention this because Ron usually supports omitting struct names. But if we require including the struct name for this syntax, it may be possible.

Do you want to submit a PR for this feature? This syntax addition would likely be feature-gated. Enabling the feature during serialisation would use the new syntax for structs (and struct-like enum variants), and enabling it during deserialisation would allow using the new syntax.

My guess is that it may be as simple as modifying the current deserialize_struct impl (and its variant counterpart). If the struct name was found, it could peek for an opening curly brace and then expect a closing curly brace later, otherwise it would use the current parsing rules. The logic for guessing struct types in deserialize_any would also need to be amended (this case would actually be quite simple to handle).

PR #551, opened 2024-11-02
Rusty struct

juntyr:
This is just a quick attempt at getting something going for #550. Anyone (e.g. @HKalbasi) is free to pick up my branch and work on this further, I might not have time for more for a while.

Should the extension just allow or require braced structs during deserialisation?

* [ ] Add tests, especially for the newtype interaction
* [ ] Add custom error variants / messages


src/de/mod.rs
```diff
@@ -241,7 +249,38 @@ impl<'de> Deserializer<'de> {
     where
         V: Visitor<'de>,
     {
-        if self.newtype_variant || self.parser.consume_char('(') {
+        if self.extensions().contains(Extensions::BRACED_STRUCTS) {
```
juntyr, line 252: This separate if statement should probably refactored to duplicate less code with the else-if branch

juntyr:
Thinking about this some more and the challenges of this implementation draft ... I think this should not be an extension but instead a pretty printing option

- this would allow all parsers to parse both the parenthesised and the braced syntax
- it would simplify the parser because it would not need to suddenly make previously allowed cases fail
- it would not require changes to RawValue
- as a downside, it would not be possible (immediately) to only parse code that follows the new syntax

PR #554, opened 2024-12-11
Add support for serializing hex strings without quotes

aljen:
## Problem

Currently, RON supports parsing hex numbers in both quoted and unquoted form:

```ron
number: 0x1234    // works fine
number: "0x1234"  // also works
```

However, when serializing such values, they are always wrapped in quotes:

```rust
pub fn as_hex_u16<S>(value: &u16, serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    serializer.serialize_str(&format!("0x{:04X}", value))
}

#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
pub struct AddressRange {
    #[serde(serialize_with = "as_hex_u16")]
    pub start: u16,
    #[serde(serialize_with = "as_hex_u16")]
    pub end: u16,
    pub length: u16,
}
```

```ron
start: "0x1234",
end: "0x1235"
length: 2
```

This creates an inconsistency between parsing and serialization capabilities, and makes the serialized output less elegant, especially in configuration files where hex numbers are commonly used (e.g., for colors, flags, or memory addresses).

## Solution

This PR adds a new configuration option `hex_as_raw` (default: `true`) to `PrettyConfig` that allows hex strings to be serialized without quotes. When enabled:

```rust
// Input
let val = "0x1234";
let config = PrettyConfig::new().hex_as_raw(true);
to_string_pretty(&val, config)?;
```

```ron
# Output
number: 0x1234
```

Number is serialised clean and consistent with parsing capabilities.

The implementation includes:

- New `hex_as_raw` configuration option
- Helper function to validate hex string format
- Comprehensive test suite covering various cases
- Full backward compatibility (can be disabled via config)

## Benefits

1. **Consistency**: Serialization now matches RON's existing parsing capabilities
2. **Readability**: Hex numbers without quotes are more readable and familiar
3. **Configurability**: The feature can be enabled/disabled as needed
4. **Safety**: Only valid hex strings (`0x[0-9a-fA-F]+`) are serialized without quotes

This enhancement makes RON's handling of hex numbers more elegant and consistent :)

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -111,6 +111,8 @@ pub struct PrettyConfig {
     pub compact_maps: bool,
     /// Enable explicit number type suffixes like `1u16`
     pub number_suffixes: bool,
+    /// Whether to serialize hex strings (0x...) without quotes
+    pub hex_as_raw: bool,
     /// Additional path-based field metadata to serialize
     pub path_meta: Option<path_meta::Field>,
 }
@@ -341,6 +343,27 @@ impl PrettyConfig {
 
         self
     }
+
+    /// Configures whether hex strings should be serialized without quotes.
+    ///
+    /// When `true`, the hex string `0x1234` will serialize to
+    /// ```ignore
+    /// 0x1234
+    /// # ;
+    /// ```
+    /// When `false`, the hex string `0x1234` will serialize to
+    /// ```ignore
+    /// "0x1234"
+    /// # ;
+    /// ```
+    ///
+    /// Default: `true`
+    #[must_use]
+    pub fn hex_as_raw(mut self, hex_as_raw: bool) -> Self {
+        self.hex_as_raw = hex_as_raw;
+
+        self
+    }
 }
 
 impl Default for PrettyConfig {
@@ -363,6 +386,7 @@ impl Default for PrettyConfig {
             compact_structs: false,
             compact_maps: false,
             number_suffixes: false,
+            hex_as_raw: true,
             path_meta: None,
         }
     }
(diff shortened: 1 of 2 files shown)


src/ser/mod.rs
```diff
@@ -489,6 +522,12 @@ impl<W: fmt::Write> Serializer<W> {
             .map_or(true, |(ref config, _)| config.escape_strings)
     }
 
+    fn hex_as_raw(&self) -> bool {
+        self.pretty
+            .as_ref()
+            .map_or(true, |(ref config, _)| config.hex_as_raw)
```
juntyr, line 528: This implies that the `hex_as_raw` option is enabled when no pretty config is given. Same with the default in the `PrettyConfig` above. Usually pretty config options are opt-in, so I'd like to hear some justification for being opt-out.

juntyr:
I didn't know we supported parsing integers from hex strings ... can you add a test for that?

juntyr:
I'm not sure if I'm in favor of this proposed change so far. If the goal is to be able to serialize all integers in binary, octal, or hex notation, I'd happily accept a proposal for that.

If you only want to serialize specific integers as hex, there is a (hacky) workaround that should work already. You could define a serde `serialize_with` function that first formats the number as a hex string, then constructs a `RawValue` from it, and then serializes that. It's not very efficient but it would already work in user code.

juntyr:
@aljen Thank you for submitting the PR! I've left some thoughts above :)

src/ser/mod.rs
aljen: Yeah, good idea, I changed it so now it defaults to false

aljen:
> I didn't know we supported parsing integers from hex strings ... can you add a test for that?

You're right, t was late, I was thinking about parsing hex value as an integer, not a string - fixed in description, sorry for confusion =)

Sorry for long time to reply, I was on holiday vacations :)

juntyr:
No worries, happy New Year!

Do you want this PR to go in the suggested direction of serialising all integers as hex / octal / binary? It would require a small rewrite in this case.

If it's just about serialising a few values in hex, have you tried out the hack using raw values?

Issue #496, opened 2023-09-03
Upcoming v0.9 release


juntyr:
I want to publish v0.9 within the next 1-2 weeks.

While I originally planned to focus on round-tripping `Value`s in v0.9, instead my focus shifted to improving the format by supporting more Rusty syntax (and serde attributes). As part of these changes, we now have much more tests, higher code coverage, and a pretty good fuzzer that keeps uncovering obscure edge cases (in the future I want to extend it to also support fuzzing serde attributes).

I now want to give the fuzzer and our users some time to find any bugs and for me to fix them. Once we've had e.g. one week without new fuzzer issues, I'll publish v0.9.

Afterwards, I will probably be a bit less active again and only fix bugs for a while, before I then shift focus to improving `Value` to become closer to an AST that can roundtrip.

- [x] Fix the byte literal grammar to include the `b'` prefix and `'` suffix + define ASCII

NiklasEi:
Just dropping in to say thank you for all your work :heart: 
I am very happy that #217 was fixed and can verify that my usecase works with latest main :slightly_smiling_face:

juntyr:
> Just dropping in to say thank you for all your work :heart: 
> 
> I am very happy that #217 was fixed and can verify that my usecase works with latest main :slightly_smiling_face: 

Thank you @NiklasEi for your kind message!

juntyr:
I wanted to give a brief update since v0.9 may be delayed by a few more days. #502 has consumed quite some time, since I want to update our fuzzer to cover all tricky serde attributes so that we can at least know what exotic combinations we do not support (see #503 for a start to expanding the restrictions section in the README). I'm almost done with updating the fuzzer, so hopefully it shouldn't take too much longer :)

juntyr:
The 60-commit fuzzer update PR is finally merged. If the fuzzer doesn't find any non-timeout or non-OOM bugs (which are related to huge arrays and which I'm not quite sure yet how to fix), I will now *finally* publish v0.9 in the coming days, now with limitations of the new serde attribute supper properly investigated.

juntyr:
I've unfortunately had to push back the release as I am an expecting parent and did not have the time to fix some fuzzer-identified (non-security) issues - I hope I will be able to get back to ron soon though :)

shenjackyuanjie:
a tiny ping, how is your child going?
are they happy? :)

juntyr:
> a tiny ping, how is your child going? are they happy? :)

[...]

PR #559, opened 2025-03-18
Add `flatten` support disclaimer

naseschwarz:
`flatten` has proven challenging to implement. In order to enable releasing 0.9 without having documented all defects, which are mostly due to serde, add an appropriate disclaimer.

See https://github.com/ron-rs/ron/issues/496

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ Furthermore, serde imposes the following restrictions for data to roundtrip:
 - internally tagged newtype variants, which are `#[serde(flatten)]`ed together with other fields, must not contain:
   - a unit or unit struct or an untagged unit variant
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -217,6 +217,8 @@ Furthermore, serde imposes the following restrictions for data to roundtrip:
 - internally tagged newtype variants, which are `#[serde(flatten)]`ed together with other fields, must not contain:
   - a unit or unit struct or an untagged unit variant
 
+While RON offers a best-effort implementation for `#[serde(flatten)]`, it may be unsupported in further cases and combinations not listed above. These limitations stem primarily from Serde rather than RON. Enumerating all such cases based on Serde's behavior is nontrivial, so the lists above are not exhaustive.
```
juntyr, line 220:
```suggestion
While RON offers a best-effort implementation for `#[serde(flatten)]`, it may be unsupported in further cases and combinations not listed above. These limitations stem primarily from serde rather than RON. Enumerating all such cases based on serde's behavior is nontrivial, so the lists above are not exhaustive.
```

juntyr:
Thanks @naseschwarz!

PR #564, opened 2025-03-20
Make Number non-exhaustive

### LINKED ISSUE #563: Enabling `integer128` feature will break code
Because [features get unified](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), the enabling of a feature should not break other packages in the dependency tree that did not enable a feature.

If one package in the dependency tree does not enable `integer128` and exhaustively destructures `Value`, then something else enabling `integer128` in the dependency tree will break that first package because new enum variants will be added that won't be taken care of.

juntyr:
Fixes #563 

### DIFF
--- a/src/value/number.rs
+++ b/src/value/number.rs
@@ -5,8 +5,35 @@ use std::{
 
 use serde::{de::Visitor, Serialize, Serializer};
 
-/// A wrapper for any numeric primitive type in Rust
+/// A wrapper for any numeric primitive type in Rust.
+///
+/// Some varints of the `Number` enum are enabled by features:
+/// - `Number::I128` and `Number::U128` by the `integer128` feature
+///
+/// To ensure that feature unification does not break `match`ing over `Number`,
+/// the `Number` enum is non-exhaustive.
+///
+/// <details>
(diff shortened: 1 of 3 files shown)


src/value/number.rs
```diff
@@ -22,6 +48,14 @@ pub enum Number {
     U128(u128),
     F32(F32),
     F64(F64),
+    #[cfg(not(doc))]
+    #[allow(private_interfaces)]
+    __NonExhaustive(private::Never),
```
epage, line 53: Unsure whats going on but this didn't make the enum exhaustive.  I wonder if the compiler recognized that this variant cannot be constructed so therefore it doesn't need to be matched?

juntyr: Uff, maybe this comes from some of the newer uninhabited type matching ...

Since `private::Never` is private, we can make the type inhabited but not constructible. Would changing this be a breaking change?

epage: Currently, people can exhaustively match on `Number`.  Making it actually non-exhaustive would then be a breaking change.

juntyr: We’ll definitely need a compile-fail test then to ensure there are no more slip ups - I’m sorry that the first fix didn’t catch this.

PR #566, opened 2025-03-23
Further deprecate base64 byte strings by removing parsing support outside the error path

juntyr:


juntyr:
I think I would like to delay this change for another release, since it's looking like we'll need a quick v0.10 release for #564 and the deprecation will need time to be picked up

PR #567, opened 2025-04-26
Add `no_std` Support

bushrat011899:
# Objective

I'm working on `no_std` support for several crates across the Rust ecosystem, and was surprised to see `ron` is one such crate! Having such support would be helpful for the Bevy game engine, as we use `ron` for metadata files. Porting our asset systems to `no_std` will require `ron` to be `no_std` as a prerequisite.

## Solution

- Added `std` feature. `std` enables the usage of the `std` crate.
- Adjusted the `SERDE_FLATTEN_CANARY` test in `src/de/mod.rs` to use a `String` rather than a byte array. This may require more investigation. I believe my fix is correct, but may over-allocate.
- Added `alloc_instead_of_core`, `std_instead_of_alloc`, and `std_instead_of_core` lints to help with maintaining `no_std` support.
- Feature gated methods relying on `std::io` behind `std`.
- Switched to a `BTreeMap` as the inner map for the `Fields` type when `std` is not active. This may not need to be feature gated and could be used in both circumstances, but I have not tested to confirm a `BTreeMap` is equivalently performant.
- Updated the usage of `IndexMap` to explicitly use the `std` `RandomState` to highlight how the `indexmap` feature could be made `no_std` in the future (requires including an alternative hasher).
- Adjusted imports to refer to `core` and `alloc` where required.
- Removed `std` features from dependencies.
- The `f32::fract` and`f64::fract` methods are not available in `no_std`. However, they can be accessed using `x % 1.` instead.
- Import `std::error::Error` from `serde::ser::StdError`. `serde` uses a `build.rs` script to work out whether it can use `std`, `core`, or if it must define its own `Error` trait, and publicly exports it. We can use that instead of needing to determine `core` vs `std` ourselves. 

## Notes

* This is my first time contributing to this project, please let me know if there's anything I can do to help!


juntyr:
Thank you for working on this feature!

I think the next step would be to fix CI by ensuring only the right feature combinations are tested. In particular, the core error feature should be run with the 1.77.

bushrat011899:
Damn, thought I'd caught all the CI issues already. I'll do another pass tonight and try and fix those issues too.

.github/workflows/ci.yaml
```diff
@@ -76,19 +76,25 @@ jobs:
       - run: cargo test --features integer128
       - run: cargo test --features indexmap
       - run: cargo test --all-features
+        if: ${{ matrix.rust != '1.64.0' }}
```
juntyr, line 79: Can you also add an explicit test for `--no-default-features` (all versions) and a note that `cargo test` tests the `std` feature?

juntyr: And an explicit test for `core_error` that doesn't run on `1.64`?

.github/workflows/ci.yaml
```diff
@@ -97,10 +103,12 @@ jobs:
           cargo update -p serde_bytes --precise 0.11.16
           cargo update -p indexmap --precise 2.5.0
           cargo update -p once_cell --precise 1.20.3
+        if: ${{ matrix.rust == '1.64.0' }}
       - run: cargo clippy -- -D warnings -A unknown-lints
       - run: cargo clippy --features integer128 -- -D warnings -A unknown-lints
       - run: cargo clippy --features indexmap -- -D warnings -A unknown-lints
       - run: cargo clippy --all-features -- -D warnings -A unknown-lints
```
juntyr, line 110: Same for the clippy checks

juntyr: And could you add a `cargo check` step as well?

src/de/mod.rs
```diff
@@ -670,25 +676,20 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
         self.newtype_variant = false;
 
-        let mut canary_buffer = [0u8; SERDE_FLATTEN_CANARY.len()];
-        std::mem::drop(write!(
-            canary_buffer.as_mut(),
-            "{}",
-            VisitorExpecting(&visitor)
-        ));
-        let terminator = if canary_buffer == SERDE_FLATTEN_CANARY {
+        let serde_flatten_canary = VisitorExpecting(&visitor)
```
juntyr, line 688: I don't like that we have to heap allocate now ... could we leave a TODO or FIXME comment to come back to this later?

src/ser/mod.rs
```diff
@@ -736,7 +737,7 @@ impl<'a, W: fmt::Write> ser::Serializer for &'a mut Serializer<W> {
 
         write!(self.output, "{}", v)?;
 
-        if v.fract() == 0.0 {
+        if v % 1. == 0.0 {
```
juntyr, line 740: Is this always the same, e.g. for negative numbers? Could you please add a test?

juntyr:
As you might see, clippy on >1.64 flags lots of things that we cannot use on 1.64. So while you're free to fix some of them, we only run clippy on the MSRV for this reason

src/ser/mod.rs
bushrat011899: Based on [`num-traits`](https://docs.rs/num-traits/latest/src/num_traits/float.rs.html#459-465), this is a valid implementation of `fract`. For the sake of simplicity, I'll just add a comment linking to this implementation.

src/de/mod.rs
bushrat011899: Agreed. I'm not familiar enough with how `Display::fmt` works to know immediately if there's an easy way to avoid allocation here. I've left a TODO as requested.

.github/workflows/ci.yaml
bushrat011899: See above for strategy testing

.github/workflows/ci.yaml
bushrat011899: Marking as resolved, but please let me know if you're unhappy with my choice here!

src/ser/path_meta.rs
```diff
@@ -98,10 +98,16 @@
...
-use std::collections::HashMap;
+use alloc::string::String;
 
 use serde_derive::{Deserialize, Serialize};
 
+#[cfg(feature = "std")]
+use std::collections::HashMap as FieldsInner;
+
+#[cfg(not(feature = "std"))]
+use alloc::collections::BTreeMap as FieldsInner;
```
bushrat011899, line 109: Future PR: It might be possible ([or even preferable](https://users.rust-lang.org/t/hashmap-vs-btreemap/13804)) to always use `BTreeMap`, removing this feature gated behaviour. Needs to be investigated with benchmarks. Since it's internal usage only, that can be done in a minor release.

src/value/map.rs
```diff
@@ -20,9 +20,9 @@ use super::Value;
 pub struct Map(pub(crate) MapInner);
 
 #[cfg(not(feature = "indexmap"))]
-type MapInner = std::collections::BTreeMap<Value, Value>;
+type MapInner = alloc::collections::BTreeMap<Value, Value>;
 #[cfg(feature = "indexmap")]
-type MapInner = indexmap::IndexMap<Value, Value>;
+type MapInner = indexmap::IndexMap<Value, Value, std::collections::hash_map::RandomState>;
```
bushrat011899, line 25: Future PR: All that's required for the `indexmap` feature to work on `no_std` is a suitable hashing algorithm to be provided. This would require a new dependency (e.g., `foldhash`, `rustc_hash`, etc.) so I'm punting it to a future PR rather than adding controversy here.

src/lib.rs
```diff
@@ -12,8 +12,17 @@
...
 #![allow(clippy::missing_errors_doc)] // FIXME
+#![warn(clippy::alloc_instead_of_core)]
+#![warn(clippy::std_instead_of_alloc)]
+#![warn(clippy::std_instead_of_core)]
 #![doc = include_str!("../README.md")]
 #![doc(html_root_url = "https://docs.rs/ron/0.10.1")]
+#![no_std]
+
+#[cfg(feature = "std")]
+extern crate std;
```
bushrat011899, line 23: This might look verbose compared to:
```rust
#![cfg_attr(not(feature = "std", no_std))]
```
But it avoids changing the implicit prelude with/out the `std` feature, which is very frustrating to develop in IME.

.github/workflows/ci.yaml
```diff
@@ -39,15 +39,23 @@ jobs:
...
+        rust: [1.64.0, 1.81.0, stable, nightly]
         lock: ["Cargo.lock", "Cargo.lock.min"]
+        integer128: [true, false]
+        std: [true, false]
+        indexmap: [true, false]
+        core_error: [true, false]
+        exclude:
+          # Feature core_error increases MSRV to 1.81.0
+          - rust: 1.64.0
+            core_error: true
```
bushrat011899, line 58: This will test every combination of features, but exclude any combinations where `core_error` is enabled on Rust 1.64. I think this should help avoid edge-case feature issues.

juntyr: Nice, thank you!

src/lib.rs
juntyr: I didn't know about this strategy, thank you for using it

src/value/map.rs
juntyr: I would prefer to support indexmap on no-std earlier rather than later. However, I agree that the choice of a hash is non-trivial. If you have a preference, please feel free to open a follow-up PR

bushrat011899:
Ok looks like it's just an issue with `serde` and `core::error::Error` I need to diagnose.

bushrat011899:
> Nit: could the CI matrix case names be changed so that instead of e.g. false, true, false, true we'd see the features used?

I'd love to do that. I think it's controlled by `run-name`, but I'll need to experiment to confirm (GitHub Actions are still pretty new to me!)

bushrat011899:
Ok I _think_ the `Check` tasks should have pretty names now, and the errors found in fuzzing and testing should be resolved (both require either the `std` or `core_error` features)

juntyr:
Thanks @bushrat011899 for all of your work on this!

I was wondering, now that the feature matrix is simpler again, if we could use cargo hack instead of a GitHub matrix to make use of caching and not spawn so many jobs.

Cargo.toml
```diff
@@ -47,3 +49,27 @@ bytes = { version = "1.3", default-features = false, features = ["serde"] }
...
+name = "encode_file"
+required-features = ["std"]
+
+[[example]]
+name = "encode"
+required-features = []
+
+[[example]]
+name = "transcode"
+required-features = []
```
juntyr, line 75:
```suggestion
required-features = []

```

juntyr (approved):

juntyr:
> Thanks @bushrat011899 for all of your work on this!
> 
> 
> 
> I was wondering, now that the feature matrix is simpler again, if we could use cargo hack instead of a GitHub matrix to make use of caching and not spawn so many jobs.

Ping @bushrat011899

bushrat011899:
> I was wondering, now that the feature matrix is simpler again, if we could use cargo hack instead of a GitHub matrix to make use of caching and not spawn so many jobs.

Sorry about the delayed response! Got distracted and forgot to come back to this, thanks for the ping! I've updated the CI task to use `cargo-hack` and remove the somewhat sketchy feature flag matrix stuff. Should be faster and easier to test locally too!

juntyr:
Thanks @bushrat011899! Finally, could you also write the CHANGELOG entry and then I'll merge your PR :)

bushrat011899:
> Thanks @bushrat011899! Finally, could you also write the CHANGELOG entry and then I'll merge your PR :)

Done! I've listed it as a breaking change, but it technically _isn't_ since nothing has changed when default features are enabled. I explicitly included all items which are now gated behind `std` just to make migration as easy as possible, but it's pretty obvious what's missing (things that use `std::io` in their public API).

juntyr:
Thank you so much @bushrat011899 for all the work you put into this PR <3

PR #568, opened 2025-04-29
Fix (again) non-exhaustive matching on Number

juntyr:


juntyr:
@epage can you please check if this would fix the issue for you?

I'm at a conference right now so it will be a few days until I can push the change and release through.

juntyr:
@epage Would this fix to ron::Number fix the feature unification and non-exhaustive matching for you?

epage:
I don't have time at the moment to test this.  Hopefully.  Still not thrilled with the direction but that is your choice.

PR #569, opened 2025-05-03
SpannedError: Store error span

pfnsec:

Hi! We're working on basic language server support for RON, both for the core syntax and as a small lib for easily bringing up custom language servers based on custom Rust types. In order to support this, we propose storing the error position in `SpannedError` as a start and end position, instead of a single position. This PR implements all the necessary logic for this proposal, and fixes all tests to validate correctly against the new `SpannedError` format.

Here's what our budding language server work looks like with our PR:

image
image
image
image

Note that this PR doesn't include the language server work itself.
In theory, this should make the modified unit tests more robust also, since they now validate against 2 positions instead of just 1.

I think it would be better to merge #567 before this PR if approved, and I'll go in and double check my changes against the new HEAD and make sure all of the tests are still passing.

P.S. thanks for RON, it's great!


juntyr:
Thanks @pfnsec for working on this feature! I like the general idea and mostly had some small design questions and thoughts on how we might be able to better test the spans.

juntyr:
@pfnsec #567 has now been merged

src/de/mod.rs
```diff
@@ -78,7 +78,8 @@ impl<'de> Deserializer<'de> {
 
         Err(SpannedError {
             code: err.into(),
-            position: Position::from_src_end(valid_input),
+            position_start: Position { line: 1, col: 1 },
```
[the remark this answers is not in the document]
pfnsec: Regarding precise range positions: `SpannedError::position` was previously defined as the line/col position of `Parser.cursor` at the point where the parser encountered an error. This ends up being at the end of the offending bytes, so we set `SpannedError::position_end` to this same line/col position. (Just thinking out loud...)

We compute `position_start` by storing a second cursor within the Parser, `prev_cursor`, and we set it to the old cursor value whenever we advance_bytes:
image

I think this is pretty logical in terms of the precise placement of ranges, what do you reckon?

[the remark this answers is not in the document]
juntyr: That seems fine, I was wondering more about what kind of range to use. For an error, it seems clear that the start should be inclusive. Do we want the end to be inclusive or exclusive? Perhaps the error could be represented using e.g. std::ops::Range (start..end).

pfnsec:
```
rustc-LLVM ERROR: IO failure on output stream: No space left on device
error: could not compile `ron` (test "449_tagged_enum")
```
  
  egads! :P

juntyr:
@pfnsec Perhaps you could add a cargo clean between each step of the CI check job?

juntyr:
I meant further above, lines 77, 80, and 85 where the check, clippy, and test steps are. Could you add a clean step between those?

pfnsec:
Ah, my bad. Like this?

src/de/mod.rs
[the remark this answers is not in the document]
pfnsec: Howdy! So, I was looking at using RangeInclusive for this to model the original arithmetic. However, after a while I think I have to concede that, while it is less verbose, this 2D range syntax was just sort of breaking my brain. For comparison:

image

The `position_start/position_end` syntax ends up being a lot more comprehensible, especially when checking a lot of different errors.

src/de/tests.rs
```diff
@@ -68,22 +68,25 @@ fn test_struct() {
         "NewType",
         Err(SpannedError {
             code: Error::ExpectedNamedStructLike("NewType"),
-            position: Position { line: 1, col: 8 },
+            position_start: Position { line: 1, col: 1 },
```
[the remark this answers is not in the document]
pfnsec: This seems simple enough - I have a similar helper for the language server side already. Where do you see this being used/emitted?

src/de/tests.rs
```diff
@@ -438,10 +477,21 @@ fn test_err_wrong_value() {
         "MyStruct(\n    x: true)",
         err(Error::ExpectedFloat, (2, 7), (2, 8)),
     );
+    check_error_span_inclusive::<MyStruct>(
+        "MyStruct(\n    x: true)",
+        err(Error::ExpectedFloat, (2, 7), (2, 8)),
+        " t",
```
juntyr, line 483: It took me a bit to figure this one out :)

src/de/tests.rs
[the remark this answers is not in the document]
pfnsec: tl;dr

- The helpers have been added, but only with `#[cfg(test)]`.
- They're test-only because I wasn't sure if `unicode-segmentation` (needed for properly selecting spans) was too heavy to add to main dependencies, so it's only in dev-dependencies.
- All unit tests that do non-trivial SpannedError checks now have substring checking as well with the new helpers.
- The new helpers have unit testing to make sure the Position and Span calculations work with, E.G. multibyte, weird unicode jazz, CRLF line endings, etc.
- Integration tests can't use them yet, because they run without `#[cfg(test)]`. To allow integration tests to use them, we'd have to add unicode-segmentation to the main dependencies. I wanted to leave that call up to you, since it's a little beefy at 104 KiB.

For reference:
| Crate                | Size     |
|----------------------|----------|
| ron                  | 111 KiB  |
| indexmap             | 89.1 KiB |
| unicode-ident        | 46.6 KiB |
| serde                | 77.1 KiB |
| unicode_segmentation | 104 KiB  |

src/de/tests.rs
```diff
@@ -699,7 +788,7 @@ fn test_leading_whitespace() {
     check_from_str_bytes_reader("  EmptyStruct1", Ok(EmptyStruct1));
 }
 
-fn check_from_str_bytes_reader<T: serde::de::DeserializeOwned + PartialEq + core::fmt::Debug>(
+pub fn check_from_str_bytes_reader<T: serde::de::DeserializeOwned + PartialEq + core::fmt::Debug>(
```
juntyr, line 791: Can you verify that this function cannot be reached in user-code?

src/de/tests.rs
```diff
@@ -716,6 +805,63 @@ fn check_from_str_bytes_reader<T: serde::de::DeserializeOwned + PartialEq + core
     }
 }
 
+/// Given a string `ron`, a SpannedResult, and a substring, verify that trying to parse `ron` results in an error
```
juntyr, line 808: [`SpannedResult`]

src/error.rs
```diff
@@ -312,20 +312,104 @@ impl Position {
     }
 }
 
+#[cfg(test)]
+impl Position {
+    /// Given a Position and a string, return the 0-indexed grapheme index into the
+    /// string at that position, or None if the Position is out of bounds of the string.
```
juntyr, line 318: [`None`]

src/error.rs
```diff
@@ -312,20 +312,104 @@ impl Position {
...
-    /// Start position of the parser before it encountered the error.
     pub start: Position,
-    /// End position (inclusive) of the parser when it encountered the error.
     pub end: Position,
 }
 
+#[cfg(test)]
+impl Span {
+    /// Given a Span and a string, form the resulting string selected exclusively (as in \[start..end\]) by the Span
+    /// or None if the span is out of bounds of the string at either end.
```
juntyr, line 385: [`None`]

src/error.rs
```diff
@@ -312,20 +312,104 @@ impl Position {
...
+        if let (Some(start), Some(end)) = (self.start.grapheme_index(s), self.end.grapheme_index(s))
+        {
+            Some(s.graphemes(true).collect::<Vec<&str>>()[start..end].concat())
+        } else {
+            None
+        }
+    }
+
+    /// Given a Span and a string, form the resulting string selected inclusively (as in \[start..=end\]) by the Span
+    /// or None if the span is out of bounds of the string at either end.
```
juntyr, line 399: [`None`]

src/de/tests.rs
juntyr: (this needs no action)

src/de/tests.rs
```diff
@@ -508,6 +570,18 @@ fn forgot_apostrophes() {
...
+    check_error_span_exclusive::<(i32, String)>(
+        "(4, \"Hello)",
+        Err(SpannedError {
+            code: Error::ExpectedStringEnd,
+            span: Span {
+                start: Position { line: 1, col: 5 },
+                end: Position { line: 1, col: 6 },
+            },
+        }),
+        "\""
```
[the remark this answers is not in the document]
juntyr: (this needs no action)

juntyr:
Functionality-wise, I'm happy with the changes and so now we just need to do a bit of cleanup

src/de/tests.rs
pfnsec: Oops, no need for this to be pub. Changed it back.

pfnsec:
Hopefully that should fix all the lints. As for the additional span tests you mentioned, we'd have to include `unicode-segmentation` as a dependency proper as mentioned above. It would be an easy addition if you're fine with the extra dep.

juntyr:
> Hopefully that should fix all the lints. As for the additional span tests you mentioned, we'd have to include `unicode-segmentation` as a dependency proper as mentioned above. It would be an easy addition if you're fine with the extra dep.

Since we'd only use it in the tests, would it still not just be a test dependency? If that doesn't work, maybe we could move the span text extraction code into its own module and add the module to those tests as well (using an explicit module path attribute)? It would be a bit of a hack but should be good enough for the three files.

pfnsec:
Oh, now I follow, you meant to say doing something like

```rust
[#path = "../src/util/span_substr.rs"]
pub mod span_substr;
```

, right? I'll take a stab at it tomorrow.

juntyr:
> Oh, now I follow, you mean doing something like
> 
> ```rust
> [#path = "../src/util/span_substr.rs"]
> pub mod span_substr;
> ```
> 
> , right? I'll take a stab at it tomorrow.

Yes, if that doesn’t work I’ve opened #571 and we can do it in a separate PR - if it does we can close the issue immediately :)

pfnsec:
Alas, it doesn't work - by importing it that way with an explicit path we end up implementing a fn for the foreign types `ron::de::Position` and `ron::de::Span`, along with other crate:: import path mismatches. I've split it off into a separate module to do the optional-feature plumbing mentioned above. I think it may be the cleanest option.

pfnsec:
Hopefully this is clean enough for our liking :) 
`unicode-segmentation` is retained as a dev dependency for unit tests. It's also `no-std` friendly, so it shouldn't interfere with the work there. The existing error checking asserts are retained as well along with the feature-gated span checks.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ default = ["std"]
 integer128 = []
 std = ["serde/std"]
 indexmap = ["std", "dep:indexmap"]
+span-substring-test = ["unicode-segmentation"]
```
juntyr, line 27: Can you add a comment here that this feature is internal and should not be used in user code (maybe even prepend "internal-" to the name?)

pfnsec:
Hold up, I think I got my wires crossed here with ron::util mod layout. Will fix in the morning...

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### API Changes
 
+- Breaking: `SpannedError` now stores the full error span from `span: Span { start: Position, end: Position }`, to facilitate, E.G., language servers highlighting syntax errors.
```
juntyr, line 11:
```suggestion
- Breaking: `SpannedError` now stores the full error span in span: Span { start: Position, end: Position }`, to facilitate, e.g., language servers highlighting of syntax errors.
```

juntyr:
@pfnsec I'm happy to merge now unless there's anything else from your end?

pfnsec:
Nope, nothing else on my end! Thanks :D

juntyr:
Thanks @pfnsec for all of your work on this feature!

PR #578, opened 2025-07-25
make value roundtrip

wiiznokes:
previous pr: #553 

todo:
- [ ] add more test cases (currently, only Enum::A pass)
- [ ] handle all struct type in `handle_any_struct`
- [ ] finish implementing `visit_enum`
- [ ] impl `Serialize` for `Value`
- [ ] clean up code

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -198,7 +198,7 @@ impl<'de> Deserializer<'de> {
                 TupleMode::ImpreciseTupleOrNewtype // Tuple and NewtypeOrTuple match equally
             },
         )?;
-        
+
         match (struct_type, ident) {
             (StructType::Unit, Some(ident)) if is_serde_content => {
                 // serde's Content type needs the ident for unit variants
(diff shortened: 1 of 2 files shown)


wiiznokes:
honestly, after having passed way to much time on this, i think its just not possible. 
We just need to attach a name to Tuple and Map.
To create a `Value`, we need a `Visitor`. The only method capable of this in `Visitor` is `visit_enum`.

But,
- the visitor must be generic. It doesn't know in advance the type of the enum.
- the `VariantAccess` trait have
   ```rs
   fn unit_variant(self) -> Result<(), Self::Error>;
   fn newtype_variant_seed<T>(self, seed: T) -> Result<T::Value, Self::Error>
   fn tuple_variant<V>(self, len: usize, visitor: V) -> Result<V::Value, Self::Error>
   fn struct_variant<V>(
         self,
         fields: &'static [&'static str],
         visitor: V,
      ) -> Result<V::Value, Self::Error>
   ``` 
   all this method take `self`, so we can't all call them to know what is the type of the enum.

So we are cooked. I wonder if a language succeed to do this. Maybe i can be proven wrong.

I think we just need another method on `VariantAccess`:
```rs
fn enum_type(&self) -> Option<EnumType>;
enum EnumType {
    Unit,
    NewType,
    Tuple,
    Struct
}
```

wiiznokes:
https://github.com/serde-rs/serde/issues/2874

wiiznokes:
Another solution could be to not implement `Deserialize` for `Value`. I don't think it's very used in his current shape anyway since it doesn't store enum anyway

PR #588, opened 2025-10-24
Add Serializer::into_inner method to extract the writer

Thomas-Mewily:
Hello, this is my first pull request :)
I will try to do my best, don’t hesitate to give me any tips on how I can improve.

# Description

This add an `into_inner()` method to `ron::ser::Serializer`, allowing users to get back the inner writer.

The name and documentation is the same as the [`serde_json::Serializer::into_inner()`](https://docs.rs/serde_json/latest/serde_json/struct.Serializer.html#method.into_inner).
(it also have the [MIT](https://spdx.org/licenses/MIT) OR [Apache-2.0](https://spdx.org/licenses/Apache-2.0) licence)

#  Motivation

A typical use case of ron looks like:

```rust
  // serializer part : serializer + buffer
  let mut buf = String::new();
  let mut ron_serializer = ron::ser::Serializer::new(&mut buf, None).unwrap();
  
  // The operation that give back the ron representation in buf
  "some structure".serialize(&mut ron_serializer).unwrap();
  println!("{buf}")
```

However, wrapping this operation into a trait + a `'static` structure is impossible without an `into_inner()`:

```rust
pub trait MarkupSerializer: Serializer {
    fn extract_markup(self) -> String; // Should be a result, but anyway this is an example
}
```

```rust
impl MarkupSerializer for &mut ron::ser::Serializer<String>
{
    fn extract_markup(self) -> String {
        panic!("impossible without an into_inner")
    }
}
```

The only way is to also store the buf in the same structure, but that can't be express in rust to my knowledge:

```rust
pub struct StringRonSerializer {
    buf: String,
    serializer: ron::Serializer<&'self mut String>, // Imaginary syntax. Reference to Self::buf . Don't compile/Not possible
}
```

So I don't think storing a ron serializer to progressively serialize somethings, them getting back the serialized string in a lifetimeless / 'static structure is not possible without an `into_inner()` method.

(My use case is to wrap a ron serializer into another serializer)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## [0.11.??] - 2025-10-24
-
 ### Additions
 - Added `into_inner()` method to `ron::ser::Serializer` to retrieve the inner writer. ([#588](https://github.com/ron-rs/ron/pull/588))
 


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.11.??] - 2025-10-24
+
```
juntyr:
```suggestion
```

PR #596, opened 2026-01-30
Add support for type and schema attributes

torkleyy:
Now we are compatible with tooling like https://github.com/ron-rs/ron2


juntyr (approved):
LGTM, thanks!

PR #597, opened 2026-02-18
fix the grammar file

wiiznokes:

draft in case i spot something else while working on a ron parser


juntyr:
Good catch! I think there unfortunately some more places where we forgot to add the `ws`

Issue #601, opened 2026-04-19
Support for number ranges


Ztry8:
What do you think about ranges like [std::ops::Range](https://doc.rust-lang.org/std/ops/struct.Range.html) in RON?
For example:
```ron
(
  value: 0..5,
  foo: -5..=2,
)
```

I think it might be just syntactic sugar and a replacement for:
```ron
(
  value: (
    start: 0,
    end: 5,
  ),
  foo: (
    start: -5,
    end: 2,
  ),
)
```

What about floats, it may be difficult and unreadable to parse due to dots:
```ron
(
    a: 0.0..0.5,
)
```

juntyr:
Syntax-wise, I see how this could be very neat. Since RON is currently tied to serde's model, and we'd want these ranges to deserialise into the range types (also once Rust switches to the new https://doc.rust-lang.org/core/range/index.html types), we'd have to make sure that the desugaring is correct, i.e. that we use the correct `start`, `end`, `last` names that the types expect.

juntyr:
Feel free to open a PR that implements the feature:
- parsing support (one issue would be that we'd need to look ahead to see that `a: 0..5` is not `a: 0`)
- a deserializer proxy that translates this into a serde mapping with the right field names
- serialising support that could look for the range types and translate them back into syntactic sugar

Ztry8:
Thanks! I have already started working on this

PR #602, opened 2026-04-25
Add support for number ranges 

Ztry8:
This PR implements #601: support for number ranges in RON.

It's just syntactic sugar for structures with fields `start, end` or `start, last`.
It supports both `std::ops` and `core::range`.

Example:
```rust
let ranges = RangeTest {
    a: 0..5,
    b: 1..=3,
    c: 0.6..4.3,
    d: 0.3..=5.7,
};

let ser = ron::to_string(&ranges).unwrap();
assert_eq!(ser, "(a:0..5,b:1..=3,c:0.6..4.3,d:0.3..=5.7)");

let de: RangeTest = ron::from_str(&ser).unwrap();
assert_eq!(de, ranges);
```


src/de/mod.rs
```diff
@@ -349,7 +349,19 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    return visitor.visit_map(RangeMapAccess::new(start, end, "end"));
+                } else if self.parser.consume_str("..") {
+                    let end = self.parser.any_number()?;
+                    return visitor.visit_map(RangeMapAccess::new(start, end, "end"));
+                }
+
+                start.visit(visitor)
+            }
             '"' | 'r' => self.deserialize_string(visitor),
             'b' if self.parser.src().starts_with("b'") => self.parser.any_number()?.visit(visitor),
```
juntyr: Byte literals should also be allowed, since they're just integers as well

src/de/mod.rs
```diff
@@ -349,7 +349,19 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             '(' => self.handle_any_struct(visitor, None),
             '[' => self.deserialize_seq(visitor),
             '{' => self.deserialize_map(visitor),
-            '0'..='9' | '+' | '-' | '.' => self.parser.any_number()?.visit(visitor),
+            '0'..='9' | '+' | '-' | '.' => {
+                let start = self.parser.any_number()?;
+
+                if self.parser.consume_str("..=") {
```
juntyr: This currently only supports bounded ranges, ranges like `a..` or `..b` could be added as a follow-up

src/de/mod.rs
```diff
@@ -715,12 +727,43 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
-        _fields: &'static [&'static str],
+        fields: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
+        if fields == ["start", "end"] || fields == ["start", "last"] {
+            if let Some(c) = self.parser.peek_char() {
+                if matches!(c, '0'..='9' | '+' | '-' | '.') {
```
juntyr: This should also allow byte literals

src/de/mod.rs
```diff
@@ -715,12 +727,43 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+
+                    let inclusive = if self.parser.consume_str("..=") {
+                        true
+                    } else if self.parser.consume_str("..") {
+                        false
+                    } else {
+                        return Err(Error::ExpectedRangeSyntax);
+                    };
+
+                    if inclusive && name == "Range" {
```
juntyr, line 798: Instead of just using the name, we could also use some of the other canaries (e.g. debug type name)

src/de/mod.rs
```diff
@@ -715,12 +727,43 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
         self,
         name: &'static str,
-        _fields: &'static [&'static str],
+        fields: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
+        if fields == ["start", "end"] || fields == ["start", "last"] {
```
juntyr: Could you match on the fields here and alias the second field so that you don't need to index `fields[1]` later

src/de/mod.rs
```diff
@@ -827,6 +870,77 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
     }
 }
 
+struct RangeMapAccess {
+    start: crate::value::Number,
+    end: crate::value::Number,
+    state: u8,
```
juntyr: Could this be an enum?

src/de/mod.rs
```diff
@@ -827,6 +870,77 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
...
+            }
+            _ => Err(Error::ExpectedDifferentLength {
+                expected: String::from("map of length 2"),
+                found: 3,
+            }),
+        }
+    }
+}
+
+struct NumberDeserializer(crate::value::Number);
```
juntyr: It might make sense to move this into the value module

src/ser/mod.rs
```diff
@@ -1050,6 +1050,12 @@ impl<'a, W: fmt::Write> ser::Serializer for &'a mut Serializer<W> {
 
     fn serialize_struct(self, name: &'static str, len: usize) -> Result<Self::SerializeStruct> {
         let old_newtype_variant = self.newtype_variant;
+        if len == 2 && (name == "Range" || name == "RangeInclusive") {
```
juntyr: Can you gate this with a new pretty config option, so that serialising with the new syntax is opt-in? Also, we should probably also use a different serialiser so that only numbers use the new syntax, since we can also only deserialise it with numbers

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,26 @@
...
+fn test_ranges() {
+    let ranges = RangeTest {
+        a: 0..5,
+        b: 1..=3,
+        c: 0.6..4.3,
+        d: 0.3..=5.7,
+    };
+
+    let ser = ron::to_string(&ranges).unwrap();
+    assert_eq!(ser, "(a:0..5,b:1..=3,c:0.6..4.3,d:0.3..=5.7)");
```
juntyr: Can you also test different integer bases, binary, octal, hex?

juntyr: Can you also add a test for deserialising inside a serde(untagged) context?

juntyr:
Thanks @Ztry8 for prototyping this feature!

juntyr:
Also, this addition will need to be documented in the grammar

Ztry8:
Thank you! I've already started working about this

Ztry8:
@juntyr, is everything fine now? What should I put into `CHANGELOG.md`? Is the `grammar.md` correct?

README.md
```diff
@@ -12,7 +12,7 @@
 
 RON is a simple readable data serialization format that looks similar to Rust syntax.
 It's designed to support all of [Serde's data model](https://serde.rs/data-model.html), so
-structs, enums, tuples, arrays, generic maps, and primitive values.
+structs, enums, tuples, arrays, generic maps, ranges and primitive values.
```
juntyr:
```suggestion
structs, enums, tuples, arrays, generic maps, ranges, and primitive values.
```

docs/grammar.md
```diff
@@ -139,6 +139,24 @@ option = "None" | option_some;
...
+## Range
+
+```ebnf
+range = range_from_to_inclusive | range_from_to_exclusive | range_from | range_to_inclusive | range_to_exclusive | range_full;
+range_from_to_inclusive = value, ws, "..=", ws, value;
+range_from_to_exclusive = value, ws, "..", ws, value;
+range_from              = value, ws, "..";
+range_to_inclusive      = "..=", ws, value;
+range_to_exclusive      = "..", ws, value;
+range_full              = "..";
```
juntyr:
```suggestion
range = range_from_to_inclusive | range_from_to_exclusive | range_from | range_to_inclusive | range_to_exclusive | range_full;
range_from_to_inclusive = number, ws, "..=", ws, number;
range_from_to_exclusive = number, ws, "..", ws, number;
range_from = number, ws, "..";
range_to_inclusive = "..=", ws, number;
range_to_exclusive = "..", ws, number;
range_full = "..";
```

docs/grammar.md
juntyr, line 56:
```suggestion
number = integer | byte | float;

integer = ["+" | "-"], unsigned, [integer_suffix];
```

docs/grammar.md
juntyr: The serialiser and parser still need to add support for full ranges - they are likely serialised as unit structs

src/de/mod.rs
```diff
@@ -349,9 +358,45 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    }
+                    let end = self.parser.any_number()?;
+                    return visitor.visit_map(RangeMapAccess::new(start, end, "end"));
+                }
+
+                start.visit(visitor)
+            }
             '"' | 'r' => self.deserialize_string(visitor),
-            'b' if self.parser.src().starts_with("b'") => self.parser.any_number()?.visit(visitor),
+            'b' if self.parser.src().starts_with("b'") => {
```
juntyr: Can you merge this arm into the other number arm above to deduplicate the inner code?

src/de/mod.rs
juntyr, line 357: What happens for a range like `inf..NaN`?

src/de/mod.rs
```diff
@@ -349,9 +358,45 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    }
+                    let end = self.parser.any_number()?;
+                    return visitor.visit_map(RangeMapAccess::new(start, end, "end"));
+                }
+
+                start.visit(visitor)
+            }
             'b' => self.deserialize_byte_buf(visitor),
             '\'' => self.deserialize_char(visitor),
             other => Err(Error::UnexpectedChar(other)),
```
juntyr, line 410: Do we need an arm here for the `..`, `..b`, and `..=b` ranges?

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
-        _fields: &'static [&'static str],
+        fields: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
+        let visitor_name = VisitorExpecting(&visitor).to_string();
+
+        if let ["start", end_field] = fields {
```
juntyr:
```suggestion
        if let ["start", end_field @ ("end" | "last")] = fields {
```

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+            if let Some(c) = self.parser.peek_char() {
+                if matches!(c, '0'..='9' | '+' | '-' | '.' | 'b')
+                    && (c != 'b' || self.parser.src().starts_with("b'"))
```
juntyr: Can you extract this condition into a helper function?

src/de/mod.rs
```diff
@@ -152,6 +153,14 @@ macro_rules! guard_recursion {
...
 }
 
+struct VisitorExpecting<V>(V);
+
+impl<'de, V: Visitor<'de>> core::fmt::Display for VisitorExpecting<&'_ V> {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+        self.0.expecting(fmt)
+    }
+}
+
```
juntyr: Thank you for adding the `VisitorExpecting` code - I changed my mind, we can just use the struct `name`, there's no need to inspect the visitor - can you please remove the visitor-inspecting code again (sorry for causing this extra work)

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    if inclusive && visitor_name == "struct Range" {
+                        return Err(Error::Message(String::from(
+                            "expected `..` for Range, found `..=`",
+                        )));
+                    }
+                    if !inclusive && visitor_name == "struct RangeInclusive" {
+                        return Err(Error::Message(String::from(
+                            "expected `..=` for RangeInclusive, found `..`",
+                        )));
+                    }
```
juntyr: This is important error handling, thanks :)

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    } else {
+                        return Err(Error::ExpectedRangeSyntax);
+                    };
+
+                    if inclusive && visitor_name == "struct Range" {
+                        return Err(Error::Message(String::from(
+                            "expected `..` for Range, found `..=`",
+                        )));
+                    }
+                    if !inclusive && visitor_name == "struct RangeInclusive" {
```
juntyr:
```suggestion
                    // check for mismatch of `..` with `RangeInclusive { start, end }` (legacy) or `RangeInclusive { start, last }` (new)
                    if !inclusive && name == "RangeInclusive" {
```

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+
+                    let inclusive = if self.parser.consume_str("..=") {
+                        true
+                    } else if self.parser.consume_str("..") {
+                        false
+                    } else {
+                        return Err(Error::ExpectedRangeSyntax);
+                    };
+
+                    if inclusive && visitor_name == "struct Range" {
```
juntyr:
```suggestion
                    // check for mismatch of `..=` with `Range { start, end }`
                    if inclusive && name == "Range" || last_name == "end") {
```

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                        )));
+                    }
+
+                    let end = self.parser.any_number()?;
+                    return visitor.visit_map(RangeMapAccess::new(start, end, end_field));
+                }
+            }
+        }
+
+        if fields == ["start"] && (name == "RangeFrom" || visitor_name == "struct RangeFrom") {
```
juntyr:
```suggestion
        if fields == ["start"] && name == "RangeFrom" {
```

src/de/mod.rs
juntyr: Let's put this into an if branch over `if name == "Range" || name == "RangeInclusive")`

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                    };
+
+                    if inclusive && visitor_name == "struct Range" {
+                        return Err(Error::Message(String::from(
+                            "expected `..` for Range, found `..=`",
+                        )));
+                    }
+                    if !inclusive && visitor_name == "struct RangeInclusive" {
+                        return Err(Error::Message(String::from(
+                            "expected `..=` for RangeInclusive, found `..`",
```
juntyr:
```suggestion
                            "expected `..=` for `RangeInclusive`, found `..`",
```

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                        true
+                    } else if self.parser.consume_str("..") {
+                        false
+                    } else {
+                        return Err(Error::ExpectedRangeSyntax);
+                    };
+
+                    if inclusive && visitor_name == "struct Range" {
+                        return Err(Error::Message(String::from(
+                            "expected `..` for Range, found `..=`",
```
juntyr:
```suggestion
                            "expected `..` for `Range`, found `..=`",
```

src/de/mod.rs
```diff
@@ -715,12 +760,94 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                        )));
+                    } else if !self.parser.consume_str("..") {
+                        return Err(Error::ExpectedRangeSyntax);
+                    }
+                    return visitor.visit_map(RangeFromMapAccess::new(start));
+                }
+            }
+        }
+
+        if fields == ["end"]
```
juntyr, line 838: Can you please split this into one branch for `RangeTo`, which always has fields `["end"]`, and one branch for `RangeToInclusive`, which has fields `["end" | "last"]`

src/de/mod.rs
```diff
@@ -827,6 +954,130 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
...
+    fn next_key_seed<K: de::DeserializeSeed<'de>>(&mut self, seed: K) -> Result<Option<K::Value>> {
+        if self.done {
+            return Ok(None);
+        }
+        seed.deserialize(de::value::StrDeserializer::<Error>::new("start"))
+            .map(Some)
+    }
+
+    fn next_value_seed<V: de::DeserializeSeed<'de>>(&mut self, seed: V) -> Result<V::Value> {
+        self.done = true;
```
juntyr, line 1065: Can you also produce an error here if `done` is already true?

src/ser/mod.rs
```diff
@@ -120,6 +120,9 @@ pub struct PrettyConfig {
     pub number_suffixes: bool,
     /// Additional path-based field metadata to serialize
     pub path_meta: Option<path_meta::Field>,
+    /// Enable compact range syntax, e.g. `0..5` instead of `(start: 0, end: 5)`.
+    /// Only works correctly when the range bounds are numbers.
```
juntyr: We must not have a feature that produces RON that cannot be deserialised. If we detect a range during serialisation, we should switch to a new deferring serialiser that checks if we get the expected fields and that both are numbers, and only then emits the new syntax, but otherwise falls back to normal struct syntax

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,96 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(PartialEq, Deserialize, Serialize, Debug)]
+struct RangeTest {
+    a: std::ops::Range<i32>,
+    b: std::ops::RangeInclusive<i32>,
+    c: std::ops::Range<f32>,
+    d: std::ops::RangeInclusive<f32>,
+}
+
```
juntyr, line 25: Can you also test the `a..` and `..b` and `..=b` and `..` cases?

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,96 @@
...
+    assert_eq!(
+        ron::from_str::<std::ops::RangeInclusive<u8>>("b'A'..=b'Z'").unwrap(),
+        b'A'..=b'Z'
+    );
+}
+
+#[derive(PartialEq, Deserialize, Serialize, Debug)]
+#[serde(untagged)]
+enum MaybeRange {
+    Range(std::ops::Range<i32>),
```
juntyr, line 96: Can you test all range types here as well?

juntyr (commented):
Thank you for the changes, I left some further comments

Ztry8:
Thank you! l've already started to working on this

src/de/mod.rs
Ztry8: Ah, sorry, I saw your suggestion after I had already committed the same code :(

src/de/mod.rs
Ztry8: We don't need it, as the tests showed

Ztry8:
@juntyr, thanks for your help and support!

I’ve implemented all of your suggestions. Could you please check?

Sorry I didn’t commit some of your changes. 
I had already pushed similar updates before seeing your comments

juntyr:
It seems like you might need to refactor some imports from `std` to `alloc`

README.md
```diff
@@ -57,6 +57,7 @@ GameConfig( // optional struct name
 * Lists: `["abc", "def"]`
 * Structs: `( foo: 1.0, bar: ( baz: "I'm nested" ) )`
 * Maps: `{ "arbitrary": "keys", "are": "allowed" }`
+* Ranges: `3..5`, `-2..=7`, `..10`, `-15..`
```
juntyr:
```suggestion
* Ranges: `3..5`, `-2.1..=7.3`, `..-15`, `..10.5`, `..=13`, `..`
```

src/de/mod.rs
```diff
@@ -152,6 +153,10 @@ macro_rules! guard_recursion {
     }};
 }
 
+fn is_number_start(c: char, src: &str) -> bool {
```
juntyr: Maybe this could be a method of the parser?

src/de/mod.rs
```diff
@@ -329,12 +334,72 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         } else if self.parser.consume_str("()") {
             return visitor.visit_unit();
         } else if self.parser.consume_ident("inf") || self.parser.consume_ident("inff32") {
+            let start = crate::value::Number::F32(crate::value::F32(core::f32::INFINITY));
+            if self.parser.consume_str("..=") {
```
juntyr: This code structure is now repeated quite a few times - could you please factor it out into a common function that is generic over f32 vs f64

src/de/mod.rs
```diff
@@ -349,10 +414,35 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             '(' => self.handle_any_struct(visitor, None),
             '[' => self.deserialize_seq(visitor),
             '{' => self.deserialize_map(visitor),
-            '0'..='9' | '+' | '-' | '.' => self.parser.any_number()?.visit(visitor),
+            '0'..='9' | '+' | '-' | '.' | 'b' => {
+                if self.parser.check_char('b') && !self.parser.src().starts_with("b'") {
+                    return self.deserialize_byte_buf(visitor);
+                }
+
+                if self.parser.check_str("..") {
```
juntyr, line 393: Can this just check the return of `consume_str`?

src/de/mod.rs
```diff
@@ -715,12 +809,118 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+
+                        let end = self.parser.any_number()?;
+                        return visitor.visit_map(RangeMapAccess::new(start, end, end_field));
+                    }
+                }
+            }
+        }
+
+        if fields == ["start"] && name == "RangeFrom" {
+            let float_ident = if self.parser.check_ident("inf") || self.parser.check_ident("inff32")
```
juntyr: Why are the inf and NaN cases needed here but not above? Could they be included in the parse_any_number code?

src/ser/mod.rs
```diff
@@ -348,6 +351,39 @@ impl PrettyConfig {
...
+    /// 0..5
+    /// # ;
+    /// ```
+    /// and `1..=3` will serialize to
+    /// ```ignore
+    /// 1..=3
+    /// # ;
+    /// ```
+    ///
+    /// Note: this option only produces correct output when the range bounds
```
juntyr: Is this comment still correct?

src/ser/mod.rs
```diff
@@ -1115,7 +1178,245 @@ impl<'a, W: fmt::Write> Compound<'a, W> {
...
+            range: Some(RangeCompound::new(inclusive, range_from)),
+        }
+    }
+}
+
+struct RangeCompound {
+    inclusive: bool,
+    range_from: bool,
+    /// Buffered fields: key → serialized string (only if numeric)
+    fields: alloc::collections::BTreeMap<&'static str, String>,
```
juntyr: Allocating here will likely be quite costly - since we know that we only have two fields at most and that we would fallback either on the first field or after serialising a number (which we can store in a `Number`), I think we should specialise for that

src/ser/mod.rs
```diff
@@ -1115,7 +1178,245 @@ impl<'a, W: fmt::Write> Compound<'a, W> {
...
+            inclusive,
+            range_from,
+            fields: alloc::collections::BTreeMap::new(),
+            fallback: false,
+        }
+    }
+
+    /// Try to serialize `value` as a number into a String buffer.
+    /// Returns `Some(s)` if numeric, `None` otherwise.
+    fn try_serialize_number<T: ?Sized + Serialize>(value: &T) -> Option<String> {
```
juntyr: I don't think we should duplicate the serialisation code to string here, especially since there are setting for serialising to other number formats, but instead to return an `Option<Number>` and then serialise that afterwards

src/ser/mod.rs
```diff
@@ -1340,6 +1641,44 @@ impl<'a, W: fmt::Write> ser::SerializeStruct for Compound<'a, W> {
...
+                    }
+                    None => {
+                        range.fallback = true;
+                    }
+                }
+                if !range.fallback {
+                    return Ok(());
+                }
+                // fallback: emit normal struct header and all buffered fields so far
+                if !self.ser.compact_structs() {
```
juntyr, line 1515: Instead of duplicating the code for fallback, maybe we could have a structure that contains the fallback fields and we then serialise that

src/ser/mod.rs
```diff
@@ -1403,6 +1742,39 @@ impl<'a, W: fmt::Write> ser::SerializeStruct for Compound<'a, W> {
     }
 
     fn end(self) -> Result<()> {
+        if let Some(ref range) = self.range {
+            if !range.fallback {
+                // All fields were numeric — emit compact syntax
+                let sep = if range.inclusive { "..=" } else { ".." };
```
juntyr: Instead of bool combinations, maybe we could use an enum

juntyr (commented):
Thank you @Ztry8 for your continued work! I've left some more comments

Ztry8:
Thanks! l've already started to working on this

Ztry8:
@juntyr, thank you for your cooperation!

I’ve implemented all of your suggestions. What do you think about them?

juntyr:
@Ztry8 Thanks for the progress! I'm a bit busy right now, can you fix CI, then I'll have the time for another full review. Thank you for all of your work!

Ztry8:
I've already started working on fixing the CI
Take your time, and sorry for rushing you

juntyr:
@Ztry8 It seems we're now getting to the tests for ranges but they are failing

Ztry8:
Yes, there are some issues that I'm currently fixing.
Sorry for the silence, I had some things to deal with related to my studies

juntyr:
@Ztry8 You shouldn’t add Cargo.lock directly - in CI we have the code that generated the lock file with maximum and minimum dependencies. If there’s a version conflict, you can look at that step to see how to fix up dependencies - often it’s resolved by using the correct version bound in Cargo.toml

Ztry8:
@juntyr, finding the root cause was quite tedious
 
The CI uses a minimal-versions lockfile (`Cargo.lock.min`) generated with `-Z minimal-versions`, which resolved `serde` to `1.0.181`

It turned out that `RangeFrom`/`RangeTo` deserialization support was added to serde in `1.0.193`, so I bumped the minimum version requirement to that

Please let me know if bumping the serde minimum version is a problem
As far as I can tell there is no other way around it, 
since the range deserialization relies on behavior introduced in that version

Sorry for the noisy chore commits 
There are quite a few of them from my debugging process,
 but the actual meaningful change is only the version bump in `Cargo.toml`. 
This should not be an issue if the PR is merged with squash

I've run CI on my fork and everything passes. 
Thanks for the feedback!!

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,194 @@
...
+    );
+
+    let de: UnclosedRangeTest = ron::from_str(&ser).unwrap();
+    assert_eq!(de, ranges);
+}
+
+#[test]
+fn test_string_range() {
+    assert!(ron::from_str::<std::ops::Range<&str>>("\"x\"..\"h\"").is_err());
+    assert!(ron::from_str::<std::ops::Range<i32>>("\"x\"..\"h\"").is_err());
```
juntyr, line 172: can you extend this test to also check the serialisation for a str range with the pretty flag turned on and off to check that it - in both cases - gives the non-compat syntax?

Ztry8:
@juntyr, thanks for the help!!
What do you think now?

src/ser/mod.rs
```diff
@@ -348,6 +350,35 @@ impl PrettyConfig {
 
         self
     }
+
+    /// Configures whether ranges should be serialized using compact syntax (`true`)
+    /// or as regular structs (`false`).
```
juntyr, line 355: Let's add a note here saying that this only works for numeric values

src/ser/mod.rs
```diff
@@ -1115,8 +1182,248 @@ impl<'a, W: fmt::Write> Compound<'a, W> {
...
+    }
+
+    /// Try to serialize `value` as a number into a String buffer.
+    /// Returns `Some(s)` if numeric, `None` otherwise.
+    #[allow(clippy::too_many_lines)]
+    fn try_serialize_number<T>(value: &T) -> Option<crate::value::Number>
+    where
+        T: ?Sized + Serialize,
+    {
+        struct NumberSerializer;
```
juntyr: Can you also move this where the number deserialiser was moved?

src/ser/mod.rs
```diff
@@ -1340,6 +1647,45 @@ impl<'a, W: fmt::Write> ser::SerializeStruct for Compound<'a, W> {
     where
         T: ?Sized + Serialize,
     {
+        if let Some(ref mut range) = self.range {
+            if !range.fallback {
+                match RangeCompound::try_serialize_number(value) {
+                    Some(s) => {
+                        if range.first.is_none() {
+                            range.first = Some(s);
```
juntyr: We should also check here that the range variant actually needs all these fields and, if it gets more fields than expected, activate the fallback

juntyr: Also, we should verify that the field key is as expected and fallback otherwise

juntyr (commented):
I'm slowly making my way through the review - now some comments on serialising

Ztry8:
@juntyr, thanks for your help and support!
I’ve implemented all of your suggestions. Could you please check?

src/ser/mod.rs
juntyr: To clarify, could you please move both the numberserializer and numberdeserializer into the value/number.rs module?

src/de/mod.rs
```diff
@@ -349,10 +379,34 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             '(' => self.handle_any_struct(visitor, None),
             '[' => self.deserialize_seq(visitor),
             '{' => self.deserialize_map(visitor),
-            '0'..='9' | '+' | '-' | '.' => self.parser.any_number()?.visit(visitor),
+            '0'..='9' | '+' | '-' | '.' | 'b' => {
+                if self.parser.check_char('b') && !self.parser.src().starts_with("b'") {
+                    return self.deserialize_byte_buf(visitor);
+                }
+
+                if self.parser.consume_str("..") {
```
juntyr: Can't this be the start of `..3` here?

src/de/mod.rs
```diff
@@ -349,10 +379,34 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                        !matches!(c, '0'..='9' | '+' | '-' | 'b')
+                            || (c == 'b' && !self.parser.src().starts_with("b'"))
+                    }) {
+                        return visitor.visit_map(RangeFromMapAccess::new(start));
+                    }
+                    let end = self.parser.any_number()?;
+                    return visitor.visit_map(RangeMapAccess::new(start, end, "end"));
+                }
+
+                start.visit(visitor)
```
juntyr: Isn't this just handle_float_range_or_value?

src/de/mod.rs
```diff
@@ -715,12 +773,95 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
         name: &'static str,
-        _fields: &'static [&'static str],
+        fields: &'static [&'static str],
         visitor: V,
     ) -> Result<V::Value>
     where
         V: Visitor<'de>,
     {
+        if name == "Range" || name == "RangeInclusive" {
+            if let ["start", end_field @ ("end" | "last")] = fields {
```
juntyr: Could it happen here that we get a mismatch between the name and end_field, i.e. should we check that we get the expected pair?

src/de/mod.rs
```diff
@@ -715,12 +773,95 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
+                return Err(Error::Message(String::from(
+                    "expected `..` for `RangeTo`, found `..=`",
+                )));
+            }
+            self.parser.consume_str("..");
+            let end = self.parser.any_number()?;
+            return visitor.visit_map(RangeToMapAccess::new(end, "end"));
+        }
+
+        if matches!(fields, ["end" | "last"])
```
juntyr, line 853: could you add a comment on why we have two possible field names (is it the old vs new fields with the new range types?)

src/de/mod.rs
```diff
@@ -827,6 +968,142 @@ impl<'a, 'de> CommaSeparated<'a, 'de> {
...
+    state: RangeMapState,
+    end_key: &'static str,
+}
+
+impl RangeMapAccess {
+    fn new(start: crate::value::Number, end: crate::value::Number, end_key: &'static str) -> Self {
+        RangeMapAccess {
+            start,
+            end,
+            state: RangeMapState::StartKey,
```
juntyr, line 993: Does really every use start in the startkey state, i.e. do we only use this for ranges that are not e.g. `..5`?

juntyr: ah, yes, we have different structs for this :)

juntyr:
@Ztry8 I left a few more comments, I think we're almost done!

Ztry8:
Thanks! l've already started to working on this

Ztry8:
@juntyr, I’m happy with the work we’ve done, thank you very much!

I’ve implemented all of your suggestions & fixed CI. 
Could you please check?

src/de/mod.rs
```diff
@@ -384,27 +384,16 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
                     return self.deserialize_byte_buf(visitor);
                 }
 
-                if self.parser.consume_str("..") {
+                if self.parser.check_str("..")
+                    && !self.parser.src()[2..].starts_with(|c: char| self.parser.is_number_start(c))
```
juntyr: I think this is not quite sufficient since there might be whitespace in between. In the parser we sometimes have this lookahead or revert pattern where we save the current cursor, parse ahead (here read the "..", skip any whitespace, then deserialise a number) and if need be reset the cursor afterwards

src/de/mod.rs
```diff
@@ -779,34 +768,37 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
-                                "expected `..=` for `RangeInclusive`, found `..`",
-                            )));
-                        }
-
-                        let end = self.parser.any_number()?;
-                        return visitor.visit_map(RangeMapAccess::new(start, end, end_field));
+        // `RangeInclusive` used `last` as field name before it was renamed to `end` in a newer Rust version
+        if (name == "Range" && fields == ["start", "end"])
+            || (name == "RangeInclusive"
+                && (fields == ["start", "end"] || fields == ["start", "last"]))
```
juntyr: You could use `matches!(fields, ["start", "end" | "last"])` here

src/ser/mod.rs
juntyr: @Ztry8 Can you please also take care of this one?

src/de/mod.rs
juntyr: Can you also please add a test for this?

Ztry8:
@juntyr, thanks for the corrections, I didn't notice them
What do you think now?

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,224 @@
...
+fn test_range_full_whitespace_lookahead() {
+    // In deserialize_any context, `..` is a unit value
+    assert!(ron::from_str::<ron::Value>("..").is_ok());
+
+    // `.. 5` with whitespace before number: old code wrongly treated this as unit in
+    // deserialize_any; with the fix it correctly sees a number follows after whitespace
+    assert!(ron::from_str::<ron::Value>(".. 5").is_err());
+
+    // Untagged enum: `.. 5` must not silently match as RangeFull/unit variant
+    assert!(ron::from_str::<MaybeRange>(".. 5").is_err());
```
juntyr, line 237: RangeFull should be a variant in MaybeRange but doesn't seem to be?

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+#[test]
+fn test_range_full_whitespace_lookahead() {
+    // In deserialize_any context, `..` is a unit value
+    assert!(ron::from_str::<ron::Value>("..").is_ok());
+
+    // `.. 5` with whitespace before number: old code wrongly treated this as unit in
+    // deserialize_any; with the fix it correctly sees a number follows after whitespace
+    assert!(ron::from_str::<ron::Value>(".. 5").is_err());
```
juntyr, line 233: Why does this fail?

tests/601_support_for_number_ranges.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    let r = ron::from_str::<std::ops::RangeInclusive<f32>>("(start:NaN,end:inf)").unwrap();
+    assert!(r.start().is_nan());
+    assert!(r.end().is_infinite());
+}
+
+#[test]
+fn test_range_full_whitespace_lookahead() {
+    // In deserialize_any context, `..` is a unit value
+    assert!(ron::from_str::<ron::Value>("..").is_ok());
```
juntyr: It would be good to check here the exact Value that is deserialised

juntyr (commented):
We're on the home stretch now - there's also a conversation from earlier on about moving the number deserialiser that I unresolved - please have a look at that as well

Ztry8:
Thanks! l've already started to working on this
Sorry about the number deserializer, I didn't understand what I needed to do at first

Ztry8:
@juntyr, many thanks!

I've improved the tests, added more clarification in the comments, and moved the number deserializer

tests/601_support_for_number_ranges.rs
```diff
@@ -212,13 +214,19 @@ fn test_inf_nan_ranges() {
...
 
-    // `.. 5` with whitespace before number: old code wrongly treated this as unit in
-    // deserialize_any; with the fix it correctly sees a number follows after whitespace
+    // `.. 5` is invalid: `..` consumes the range-full token, but `5` is trailing garbage
     assert!(ron::from_str::<ron::Value>(".. 5").is_err());
 
-    // Untagged enum: `.. 5` must not silently match as RangeFull/unit variant
+    // Untagged enum: `.. 5` must not silently match as RangeFull/unit variant,
+    // because `5` is trailing garbage after `..`
     assert!(ron::from_str::<MaybeRange>(".. 5").is_err());
```
juntyr, line 237: Could you also extend the test with the case where `.. 5` deserialises correctly, probably with a separate untagged enum?

src/value/mod.rs
```diff
@@ -255,6 +255,25 @@ impl<'a, 'de> MapAccess<'de> for MapAccessor<'a> {
     }
 }
 
+pub(crate) struct NumberDeserializer(pub(crate) Number);
```
juntyr: Can you also move this into the number module?

Let's also make sure that both the number serializer and deserializer remain private (you can split the use above into a separate pub(crate) use)

juntyr:
@Ztry8 Thanks, we're almost almost there - thank you so much for making all these final adjustments

Ztry8:
@juntyr, thank you for your help and cooperation!
I'm glad I can help make RON better, and I'm really happy with the progress we've made!

What's left to do?

tests/601_support_for_number_ranges.rs
```diff
@@ -224,6 +236,15 @@ fn test_range_full_whitespace_lookahead() {
     // because `5` is trailing garbage after `..`
     assert!(ron::from_str::<MaybeRange>(".. 5").is_err());
 
+    // Untagged enum where RangeFull comes after Value: `.. 5` is invalid RON
+    // regardless of variant order — `..` is not a valid prefix for integers
+    assert!(ron::from_str::<MaybeRangeOrValue>(".. 5").is_err());
```
juntyr, line 246: I meant to add a successful variant for `RangeTo` in addition to this new test

tests/601_support_for_number_ranges.rs
```diff
@@ -212,6 +212,18 @@ fn test_inf_nan_ranges() {
     assert!(r.end().is_infinite());
 }
 
+// Untagged enum where RangeFull comes after Value: `.. 5` should correctly
+// fall through the RangeFull attempt (trailing garbage) and match Value(5)
```
juntyr: This comment seems incorrect

Ztry8:
@juntyr, what do you think now?

tests/601_support_for_number_ranges.rs
```diff
@@ -236,6 +236,15 @@ fn test_range_full_whitespace_lookahead() {
...
 
+    assert_eq!(
+        ron::from_str::<MaybeRangeOrValue>("..5").unwrap(),
+        MaybeRangeOrValue::RangeTo(..5)
+    );
+    assert_eq!(
+        ron::from_str::<MaybeRangeOrValue>("..5").unwrap(),
+        MaybeRangeOrValue::RangeTo(..5)
+    );
+
```
juntyr, line 243: This is duplicated

Ztry8:
@juntyr, what do you think now?

juntyr:
@Ztry8 Thank you so much for your amazing work on this feature! I've now merged the PR.

Let's give the PR a few days on main to try it out, and then I'll release a new version with the feature (feel free to ping me if I forget).

Ztry8:
@juntyr, thank you for your help and cooperation!
I'm glad I can help make RON better, and I'm really happy with the progress we've made!

Ztry8:
@juntyr, has enough time passed for a new release?

juntyr:
> @juntyr, has enough time passed for a new release?

Thanks for the ping, I've now released v0.12.2

Issue #585, opened 2025-10-03
Add option for deserialising multiline strings with indentation


funnyboy-roks:
When deserialising a multiline string, an indentation of the lines is added to the string, so if a user doesn't want them to be indented, the lines must be pushed to the left edge.

This causes strings to be fairly hard to read when the field is nested:

```ron
(
    foo: (
         bar: "this
is
a
multiline
string"
    )
)
```

I propose that an option be added to the deserialiser that will support the following:

```ron
(
    foo: (
        bar: "this
              is
              a
              multiline
              string"
    )
)
```

(I think using an option is best so as to not break existing files)

I think the simplest method to do this would be to use the same logic as [`unindent`](https://docs.rs/unindent) (or even the crate itself)

thurn:
@juntyr do you know if there's a specific reason why RON doesn't support \ line continuation for strings the way Rust does? I was actually very surprised that it works differently, and that existing syntax would remove a lot of the need for this.

juntyr:
@thurn There's no reason for it; we just haven't implemented support for it. Please feel free to submit a PR that updates the parser and grammar :)

PR #605, opened 2026-06-26
Round-trip compact ranges containing inf/NaN

mahirhir:
The compact-range serializer emits `inf..2.0` / `NaN..=2.0`, but the deserializer rejects its own output, so a round-trip fails. PR #602 added float handling to `RangeFrom` but left `Range` and `RangeInclusive` unmirrored. The ~8-line fix applies the same handling to the two siblings. Round-trip verified both ways.

### DIFF
--- a/src/de/mod.rs
+++ b/src/de/mod.rs
@@ -784,7 +784,14 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
         {
             let end_field = fields[1];
             if let Some(c) = self.parser.peek_char() {
-                if self.parser.is_number_start(c) {
+                if self.parser.is_number_start(c)
+                    || self.parser.check_ident("inf")
+                    || self.parser.check_ident("inff32")
+                    || self.parser.check_ident("inff64")
+                    || self.parser.check_ident("NaN")
+                    || self.parser.check_ident("NaNf32")
+                    || self.parser.check_ident("NaNf64")
+                {
                     let start = self.parser.any_number()?;
 
                     let inclusive = if self.parser.consume_str("..=") {
--- a/tests/601_support_for_number_ranges.rs
+++ b/tests/601_support_for_number_ranges.rs
@@ -256,3 +256,53 @@ fn test_range_full_whitespace_lookahead() {
         MaybeRange::RangeFull(std::ops::RangeFull)
     );
 }
+
+#[test]
+fn test_range_inf_nan_start_bound_roundtrips() {
+    // Regression: the compact-range serializer emits `inf`/`NaN` start bounds
+    // (e.g. `inf..2.0`, `NaN..=2.0`), but the `Range`/`RangeInclusive`
+    // deserializers only entered the compact-range path when the first byte
+    // was recognised by `is_number_start` (digit / `+` / `-` / `.` / `b`).
+    // Since `inf` and `NaN` start with a letter, those inputs failed to
+    // round-trip with `ExpectedNamedStructLike`, even though `RangeFrom`
+    // (e.g. `inf..`) round-tripped fine.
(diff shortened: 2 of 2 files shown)


src/de/mod.rs
```diff
@@ -784,7 +784,14 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
             if let Some(c) = self.parser.peek_char() {
-                if self.parser.is_number_start(c) {
+                if self.parser.is_number_start(c)
+                    || self.parser.check_ident("inf")
+                    || self.parser.check_ident("inff32")
+                    || self.parser.check_ident("inff64")
+                    || self.parser.check_ident("NaN")
+                    || self.parser.check_ident("NaNf32")
+                    || self.parser.check_ident("NaNf64")
+                {
```
juntyr: Could you please factor this out into a method for the parser, since we now repeat this code a few times?

juntyr (commented):
Thanks @greymoth-jp for catching the issue and writing the fix!

src/de/mod.rs
mahirhir: Factored out as `Parser::check_any_number_start`, placed next to `any_number` because it answers whether `any_number` has something to parse at the cursor: `is_number_start` on the peeked char, or one of the six `inf`/`NaN` idents that begin with a letter. Both the `Range`/`RangeInclusive` arm and the `RangeFrom` arm call it now. The `RangeFrom` arm folds into its outer `if` since the `peek_char` binding is gone, which also clears `clippy::collapsible_if`.

Verified locally: `cargo fmt --check` clean, no new clippy findings on the changed lines, and `cargo test --test 601_support_for_number_ranges` passes 8 of 8. My local full run has pre-existing failures from a CRLF checkout on Windows in the serializer string comparisons, identical with and without this change, so I am relying on CI for the whole suite.

juntyr (approved):

juntyr:
Thanks @mahirhir, could you please also add a change log entry?

mahirhir:
Added, as `b485d55`.

```
## Unreleased

### Bug Fixes

- Fixed round-tripping of compact ranges whose bounds are `inf` or `NaN` ([#605](https://github.com/ron-rs/ron/pull/605))
```

I put it under a new `### Bug Fixes` heading rather than `### Format Changes`, since nothing about the format changed here, only that the deserializer now accepts what the serializer already emitted. Say the word if you would rather it sat under the existing heading.

One thing worth flagging: my branch predates the #585 entry that is on master's Unreleased section now, so the merge may want a hand there. Happy to merge master in if that is easier for you than resolving it.

mahirhir:
Merged master in (`69c9467`) and corrected the changelog line (`6e6bcf6`).

The conflict turned out to be only `CHANGELOG.md` — both sides had added a section under an `## Unreleased` heading the base did not have. I rebuilt it from your bytes and put Bug Fixes after the #585 Format Changes bullet, matching how 0.9.0, 0.12.0 and 0.12.1 order those two.

`src/parse.rs` auto-merged. I checked rather than assumed: `git diff upstream/master -- src/parse.rs` is exactly my 18 lines around `check_any_number_start`, and your `is_string_continuation_whitespace`, the bounded `..` search in `next_bytes_is_float`, the `find(['"','\\'])` loop and `consume_string_continuation` are all still there.

Full suite after the merge: 462 passed, 0 failed. `601_support_for_number_ranges` 8, `607_quadratic_float_parsing` 1, `609_quadratic_escaped_string_parsing` 2, `escape` 12.

I also fixed my own changelog wording. It said "whose bounds are inf or NaN", which overstates it — only the start bound was broken, since the guard was `is_number_start` and the end already went through `any_number`. It now says start bound.

The two workflow runs sit at `action_required`, which I think needs your approval to run on a fork PR.

PR #606, opened 2026-07-08
fix: correct typo in Serializer doc comment

maxtaran2010:
The doc comment on the `Serializer` struct incorrectly says "deserializing" when it should say "serializing". The `to_string` function is a serialization function, not a deserialization function.

**Before:**
\`\`\`rust
/// You can just use [`to_string`] for deserializing a value.
\`\`\`

**After:**
\`\`\`rust
/// You can just use [`to_string`] for serializing a value.
\`\`\`

### DIFF
--- a/src/ser/mod.rs
+++ b/src/ser/mod.rs
@@ -413,7 +413,7 @@ impl Default for PrettyConfig {
 
 /// The RON serializer.
 ///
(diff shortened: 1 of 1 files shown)


juntyr:
Thanks for catching the error and providing the fix @maxtaran2010!

PR #608, opened 2026-07-15
Fix quadratic float parsing in next_bytes_is_float

### LINKED ISSUE #607: Quadratic float parsing in v0.12.2: next_bytes_is_float scans to EOF per number
### Summary

Since v0.12.2, deserializing number-heavy documents is **quadratic** in the number count. It is a regression introduced by #602 (number range support, 6529e97) and it only affects self-describing deserialization (`ron::Value` / `deserialize_any`).

On a flat list of floats, v0.12.1 is linear and v0.12.2 is quadratic — a 462× slowdown at 64k numbers, and it keeps growing:

| floats | v0.12.1 | v0.12.2 | slowdown |
|--------|---------|---------|----------|
| 16 000 | 3.2 ms  | 335 ms   | 105× |
| 32 000 | 6.4 ms  | 1 397 ms | 218× |
| 64 000 | 12.2 ms | 5 645 ms | **462×** |

Doubling the input multiplies v0.12.2's time by ~4 (quadratic) vs ~2 for v0.12.1 (linear). Measurements are `--release`.

### Cause

`Parser::next_bytes_is_float` searches the **entire remaining input** for `".."`, but clamps the result to the current float-char run:

https://github.com/ron-rs/ron/blob/master/src/parse.rs#L1032-L1049

```rust
let raw_float_len = self.next_chars_while_from_len(skip, is_float_char);
// Trim at ".." to avoid treating range operators as float chars
let valid_float_len = self.src()[skip..]
    .find("..")                              // <-- scans to EOF
    .map(|i| i.min(raw_float_len))
    .map_or(raw_float_len, |i| i.min(raw_float_len));
```

Because the result is `.min(raw_float_len)` anyway, every byte scanned past `raw_float_len` is wasted. In a document that contains no `".."` at all — the common case — `find` scans all the way to EOF for **every number parsed**, giving O(numbers × input_size).

`any_number` a few lines above already does the same thing correctly, by bounding the slice before searching:

```rust
let raw_bytes = self.next_chars_while_len(is_float_char);
let src = &self.src()[..raw_bytes];
let num_bytes = src.find("..").unwrap_or(raw_bytes);
```

### Fix

Bound the search to the float-char run, matching `any_number`:

```rust
let valid_float_len = self.src()[skip..][..raw_float_len]
    .find("..")
    .map_or(raw_float_len, |i| i.min(raw_float_len));
```

This is behaviour-preserving: a match at index `>= raw_float_len` was already clamped away, and a `".."` cannot straddle the end of the run — that would require `src[raw_float_len] == '.'`, but `'.'` is a float char (`is_float_char`) and would have been part of the run.

Restores linear scaling and makes the 64k case **207× faster** (5 645 ms → 27 ms). All existing tests pass, including the seven range tests from #602.

I have a PR ready with the fix and a scaling regression test.

### Repro

```rust
// ron = "=0.12.2"
fn main() {
    for n in [8_000usize, 16_000, 32_000, 64_000] {
        let src = format!("[{}]", vec!["1234.5678"; n].join(","));
        let t = std::time::Instant::now();
        let v: ron::Value = ron::from_str(&src).unwrap();
        std::hint::black_box(&v);
        println!("{n:>6} floats: {:>8.1} ms", t.elapsed().as_secs_f64() * 1000.0);
    }
}
```

Note the typed path (`Vec<f64>`) is unaffected — it does not go through `next_bytes_is_float`.

enomado:
Fixes #607.

### Problem

`Parser::next_bytes_is_float` searches the **entire remaining input** for `".."`, but clamps the result to the current float-char run:

```rust
let raw_float_len = self.next_chars_while_from_len(skip, is_float_char);
let valid_float_len = self.src()[skip..]
    .find("..")                              // scans to EOF
    .map(|i| i.min(raw_float_len))
    .map_or(raw_float_len, |i| i.min(raw_float_len));
```

Since the result is `.min(raw_float_len)` anyway, every byte scanned past `raw_float_len` is wasted. In a document with no `".."` in it — the common case — `find` scans to EOF for **every number**, making self-describing deserialization (`ron::Value` / `deserialize_any`) O(numbers × input_size).

This regressed in v0.12.2 with #602. Same benchmark, only the version differs (`--release`):

| floats | v0.12.1 | v0.12.2 | slowdown |
|--------|---------|---------|----------|
| 16 000 | 3.2 ms  | 335 ms   | 105× |
| 32 000 | 6.4 ms  | 1 397 ms | 218× |
| 64 000 | 12.2 ms (×1.90) | 5 645 ms (×4.04) | **462×** |

### Fix

Bound the search to the float-char run — the form `any_number` already uses a few lines above:

```rust
let valid_float_len = self.src()[skip..][..raw_float_len]
    .find("..")
    .map_or(raw_float_len, |i| i.min(raw_float_len));
```

Behaviour-preserving:

- A match at index `>= raw_float_len` was already clamped to `raw_float_len`, so not finding it changes nothing.
- A `".."` cannot straddle the end of the run: that would require `src[raw_float_len] == '.'`, but `'.'` is a float char (`is_float_char`) and would have been part of the run.

The now-redundant `.map(|i| i.min(raw_float_len))` before `.map_or(...)` is dropped (`map_or` already re-applies the same `min`).

Result: linear again, **64k floats 5 645 ms → 27 ms (207×)**.

### Tests

- All 453 existing tests pass, including the seven range tests from #602.
- Adds `tests/607_quadratic_float_parsing.rs`, which asserts on **scaling** rather than absolute time so it stays meaningful on slow or noisy CI: the ratio is ~2 when linear, ~4 when quadratic, and the threshold (3.0) sits far from both. It bails out if the baseline is too small to measure.

I verified the test genuinely catches the bug: on the unfixed parser it fails with `ratio 3.90`, and it passes on the fixed one.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1036,10 +1036,20 @@ impl<'a> Parser<'a> {
                 _ => 0,
             };
             let raw_float_len = self.next_chars_while_from_len(skip, is_float_char);
-            // Trim at ".." to avoid treating range operators as float chars
-            let valid_float_len = self.src()[skip..]
+            // Trim at ".." to avoid treating range operators as float chars.
+            //
+            // Only search within the float-char run: the result is clamped to
+            // `raw_float_len` anyway, so a match at or beyond it cannot change the
+            // outcome. Searching the whole remaining input made this O(remaining)
+            // per number, i.e. quadratic in the number count for documents that
+            // contain no ".." at all (every number then scanned to EOF).
+            //
+            // A ".." cannot straddle the end of the run: that would require
+            // `src[raw_float_len] == '.'`, but '.' is a float char and would have
+            // been part of the run. `any_number` above already uses this same
+            // bounded-slice form.
+            let valid_float_len = self.src()[skip..][..raw_float_len]
                 .find("..")
-                .map(|i| i.min(raw_float_len))
                 .map_or(raw_float_len, |i| i.min(raw_float_len));
             let valid_int_len = self.next_chars_while_from_len(skip, is_int_char);
             valid_float_len > valid_int_len
--- a/tests/607_quadratic_float_parsing.rs
+++ b/tests/607_quadratic_float_parsing.rs
@@ -25,10 +25,8 @@ fn seq_of_floats(n: usize) -> String {
 fn parse_millis(n: usize) -> f64 {
     let src = seq_of_floats(n);
     let start = Instant::now();
-    // Bind (and later drop) the parsed value without `std::hint::black_box`,
-    // which is only stable since Rust 1.66 and would break the 1.64 MSRV.
-    // The parse can't be optimized away regardless: `from_str::<Value>` is a
-    // cross-crate, non-inlined call that allocates and is `.unwrap()`ed.
+    // No `black_box` (it needs Rust 1.66, MSRV is 1.64): the unwrapped
+    // cross-crate `from_str` allocates, so the parse isn't optimized away.
     let _value: ron::Value = ron::from_str(&src).unwrap();
     let elapsed = start.elapsed().as_secs_f64() * 1000.0;
     elapsed
--- a/tests/607_quadratic_float_parsing.rs
+++ b/tests/607_quadratic_float_parsing.rs
@@ -1,0 +1,60 @@
+//! Regression test: parsing number-heavy documents must stay linear.
+//!
+//! `Parser::next_bytes_is_float` used to call `find("..")` on the whole remaining
+//! input while clamping the result to the current float-char run, which made every
+//! number scan to EOF on documents containing no `..` at all — quadratic overall.
+//!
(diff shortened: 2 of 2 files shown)


juntyr:
Thanks for the catch and the fix!

juntyr:
@enomado It seems like we might need to disable the test on the 1.64

juntyr:
Or maybe it could work without the black box hint?

juntyr (approved):

enomado:
Went with your second suggestion — dropped the `black_box` hint instead of gating the test off 1.64, so it keeps running everywhere. That was the only real breakage: `std::hint::black_box` is only stable since 1.66, so the 1.64 MSRV job hit E0658 (the `stable`/`nightly` jobs were just cancelled by fail-fast). The `.unwrap()`ed cross-crate `from_str::<Value>` allocates, so the parse still isn't optimized away without the hint. CI is green now.

PR #610, opened 2026-07-15
Fix quadratic escaped string parsing in escaped_byte_buf

### LINKED ISSUE #609: Parsing strings with many escapes is quadratic (regressed in #534)
`Parser::escaped_byte_buf` calls `find('"')` once per escape. That call scans from
the cursor all the way to the **closing quote**, but the cursor only advances past
**one escape** per iteration — so a string with N escapes rescans the tail N times.
Overall: O(N²).

Unlike #607, this is not limited to `ron::Value` / `deserialize_any`: it affects any
deserialization of a string containing escapes, including the typed path.

## Repro

```rust
let src = format!("[\"{}\"]", "\\n".repeat(n));
let _: ron::Value = ron::from_str(&src).unwrap();
```

Release build, at `d0e99bc` (current master):

| n | input | time | ratio |
|---|---|---|---|
| 12 500 | 25 KB | 6.64 ms | — |
| 25 000 | 50 KB | 26.24 ms | x3.95 |
| 50 000 | 100 KB | 103.38 ms | x3.94 |
| 100 000 | 200 KB | 414.43 ms | x4.01 |
| 200 000 | 400 KB | 1652.96 ms | x3.99 |

A ratio of ~4 per doubling is quadratic (~2 would be linear). Throughput collapses to
0.2 MB/s at 400 KB, against ~77 MB/s for an ordinary RON document on the same machine.

## Root cause

Introduced by #534 ("Speed up escaped byte buf parsing"). Before that PR the loop did a
single scan for whichever of `"` / `\\` came first:

```rust
let (new_i, end_or_escape) = self
    .find_char_index(|c| matches!(c, '\\' | '"'))
    .ok_or(Error::ExpectedStringEnd)?;
```

which is linear. #534 replaced it with "find the closing quote, then look for a backslash
before it". That is genuinely faster for the common **no-escape** case — the fast path
*outside* the loop, which is fine and worth keeping — but *inside* the loop the extra
`find('"')` rescans the tail on every escape.

`git tag --contains 08c691d0` ⇒ affected releases: **v0.9.0 through v0.12.2**.

## Fix

Only the first of `"` / `\\` is actually needed; scanning just to the nearest delimiter
keeps the loop linear while leaving #534's non-escape fast path untouched.

Measured with that change: **1652.96 ms → 1.54 ms (1073x)**, ratio 3.99 → 2.02 (linear).

Happy to send a PR.

enomado:
Fixes #609.

`escaped_byte_buf` calls `find('"')` once per escape. That call scans from the cursor to
the **closing quote**, while the cursor only advances past **one escape** per iteration —
so a string with N escapes rescans the tail N times, making string parsing O(N²) in the
number of escapes.

Unlike #607, this is not limited to `ron::Value` / `deserialize_any`: it affects any
deserialization of a string containing escapes, including the typed path.

## Root cause

The quadratic scan came in with #534, which replaced a single scan for the first of
`"` / `\\`:

```rust
let (new_i, end_or_escape) = self
    .find_char_index(|c| matches!(c, '\\' | '"'))
    .ok_or(Error::ExpectedStringEnd)?;
```

with "find the closing quote, then look for a backslash before it". That is genuinely
faster for the common **no-escape** case, so this PR keeps it as the fast path *outside*
the loop. Inside the loop, though, only the first of `"` / `\\` is ever needed — scanning
just to the nearest delimiter is both sufficient and linear.

Affected releases (`git tag --contains 08c691d0`): **v0.9.0 … v0.12.2**.

## Result

`["\n\n\n…"]`, release build, at `d0e99bc`:

| input | before | after | ratio before | ratio after |
|---|---|---|---|---|
| 25 KB | 6.64 ms | 0.10 ms | — | — |
| 50 KB | 26.24 ms | 0.19 ms | x3.95 | x1.94 |
| 100 KB | 103.38 ms | 0.38 ms | x3.94 | x1.97 |
| 200 KB | 414.43 ms | 0.76 ms | x4.01 | x2.00 |
| 400 KB | 1652.96 ms | **1.54 ms** | x3.99 | **x2.02** |

**1073x** at 400 KB, and the per-doubling ratio goes from ~4 (quadratic) back to ~2
(linear). Throughput: 0.2 MB/s → 260 MB/s.

Both branches already did the same `extend_from_slice`, so the fix also makes the loop
slightly shorter.

## Tests

The regression test asserts on **scaling**, not absolute time, so it stays meaningful on
slow or noisy CI: the ratio is ~2 when linear and ~4 when quadratic, and the threshold
(3.0) sits far from both. It fails on the pre-fix parser (measured ratio 3.96) and passes
after. Same shape as the test in #608.

The existing suite (453 tests) stays green, plus the 2 new ones.

## Note on error reporting for invalid input

For an unterminated string whose invalid escape follows an escaped quote — e.g.
`"a\"b\c` — this now reports `InvalidEscape` instead of `ExpectedStringEnd`. That is what
the parser did before #534 (verified by running the pre-#534 tree), and it is arguably the
more accurate diagnostic, since there genuinely is an invalid escape.

The outer `find('"')` is deliberately left untouched, so the common case — an unterminated
string with no escapes, e.g. the `(4, "Hello)` case pinned in `de::tests::forgot_apostrophes`
— still reports `ExpectedStringEnd` exactly as today. No existing test changes behaviour.

Happy to preserve the current error exactly if you'd prefer; it costs a little extra state
to track the last found quote across iterations.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1217,17 +1217,21 @@ impl<'a> Parser<'a> {
                     },
                 }
 
-                // Checking for '"' and '\\' separately is faster than searching for both at the same time
-                let new_str_end = self.src().find('"').ok_or(Error::ExpectedStringEnd)?;
-                let new_escape = self.src()[..new_str_end].find('\\');
+                // Unlike the non-escaped case above, searching for '"' and '\\'
+                // separately is *not* sound here: `find('"')` scans all the way to
+                // the closing quote, but the cursor only advances by one escape per
+                // iteration, so a string with N escapes rescans the tail N times.
+                // Only the first of '"' / '\\' is needed, and scanning just to the
+                // nearest delimiter keeps the loop linear.
+                let next = self.src().find(['"', '\\']).ok_or(Error::ExpectedStringEnd)?;
+                s.extend_from_slice(&self.src().as_bytes()[..next]);
 
-                if let Some(new_escape) = new_escape {
-                    s.extend_from_slice(&self.src().as_bytes()[..new_escape]);
-                    i = new_escape;
+                // `next` indexes an ASCII byte, so byte indexing is valid here.
+                if self.src().as_bytes()[next] == b'\\' {
+                    i = next;
                 } else {
-                    s.extend_from_slice(&self.src().as_bytes()[..new_str_end]);
                     // Advance to the end of the string + 1 for the `"`.
-                    break Ok((ParsedByteStr::Allocated(s), new_str_end + 1));
+                    break Ok((ParsedByteStr::Allocated(s), next + 1));
                 }
             }
         } else {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1223,7 +1223,10 @@ impl<'a> Parser<'a> {
                 // iteration, so a string with N escapes rescans the tail N times.
                 // Only the first of '"' / '\\' is needed, and scanning just to the
                 // nearest delimiter keeps the loop linear.
-                let next = self.src().find(['"', '\\']).ok_or(Error::ExpectedStringEnd)?;
+                let next = self
+                    .src()
+                    .find(['"', '\\'])
+                    .ok_or(Error::ExpectedStringEnd)?;
                 s.extend_from_slice(&self.src().as_bytes()[..next]);
 
                 // `next` indexes an ASCII byte, so byte indexing is valid here.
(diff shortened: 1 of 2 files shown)


enomado:
Fixed the CI: dropped the `black_box` hint (only stable since Rust 1.66, so it broke the 1.64 MSRV job with E0658) and split the long `find(['"', '\\'])` line to satisfy rustfmt. All green now.

Issue #614, opened 2026-07-16
docs/grammar.md diverges from the parser in 6 places (string/char escapes, ranges, exponent)


enomado:
## Summary

While building a differential between a PEG reference transcribed verbatim from
[`docs/grammar.md`](https://github.com/ron-rs/ron/blob/master/docs/grammar.md)
and the actual parser, I found **6 places where the documented grammar
disagrees with what `ron` parses**. In every case the parser behaves sensibly —
it's the grammar doc that is wrong. All are present on `master`.

Method: a reference acceptor (pest) checked against
`Deserializer::from_str -> IgnoredAny::deserialize -> end()` (pure syntactic
validation of one value; `Value` can't be used as the oracle since it lacks
struct/enum/range/byte variants). The reference was then hardened with mutation
fuzzing of valid seeds, which is what surfaced the range / char-apostrophe /
exponent findings below — cases too easy to miss by hand.

Repro convention below: "ron: accept/reject" is the result of the oracle above.
Line references are from `v0.12.2` but the productions are unchanged on master.

---

### 1. `string_std` / `byte_string_std` — a lone `\` is allowed as a literal

```ebnf
string_std = "\"", { no_double_quotation_marks | string_escape }, "\"";
```
`no_double_quotation_marks` excludes only `"`, so the grammar accepts a backslash
that does not begin a valid escape as a literal character — e.g. `"a\c"`. The
parser rejects this: inside a std string a `\` must introduce a valid escape
(`parse_escape` → `InvalidEscape("Unknown escape character")`). Same for `b"…"`.

- `"a\c"` → grammar: accept, ron: **reject**
- `"a\"b\c"` → grammar: accept, ron: **reject**

**Fix:** the non-escape character class must exclude `\` as well as `"`, i.e. a
`\` always introduces a `string_escape`:
```ebnf
string_std = "\"", { no_double_quote_or_backslash | string_escape }, "\"";
```

### 2. `char` — escape set is too narrow

```ebnf
char = "'", (no_apostrophe | "\\\\" | "\\'"), "'";
```
Only `\\` and `\'` are allowed, but `char()` calls the same
`parse_escape(is_char = true)` as strings, so the full escape set works:

- `'\n'`, `'\t'`, `'\0'` → grammar: reject, ron: **accept**
- `'\x41'` → grammar: reject, ron: **accept** (single byte, must be `<= 0x7F`)
- `'\u{41}'` → grammar: reject, ron: **accept**

[...]

PR #615, opened 2026-07-16
test: grammar-conformance differential (transcribe docs/grammar.md, diff against the parser)

enomado:
## What

ron has ~450 correctness tests but nothing checks that the parser agrees with the
grammar it documents in
[`docs/grammar.md`](https://github.com/ron-rs/ron/blob/master/docs/grammar.md) —
there is no grammar-conformance test.

This adds one. `conformance/` is an isolated sub-crate — the same trick as
`fuzz/` (its own `[workspace]`), so **`pest` never enters ron's own dependency
graph or the MSRV job**. It transcribes the EBNF **verbatim** into a PEG acceptor
(`conformance/src/ron.pest`) and diffs it, input by input, against the parser. A
disagreement is either a parser bug or a grammar-doc bug.

The oracle for "does ron accept this as one syntactically-valid value?" is
`Deserializer::from_str -> IgnoredAny::deserialize -> end()`: pure syntactic
validation, whole input consumed, no data model imposed. (`Value` can't be the
oracle — it lacks struct/enum/range/byte variants and would reject or mangle them.)

## What it found

Building the differential surfaced **seven** places where `docs/grammar.md`
disagrees with the parser — in every case the parser is right and the doc is
wrong. They are filed as #614 and pinned here in `KNOWN_DOC_BUGS`:

| # | input | doc | parser | gap |
|---|-------|-----|--------|-----|
| A | `"a\c"` | accept | reject | a lone `\` must start a valid escape |
| B | `'\n'` | reject | accept | char takes the full string escape set |
| C | `"￿"` | accept | reject | `\u` requires `{ 1..=6 hex }` |
| D | `1 ..2` | accept | reject | no whitespace around range operators |
| E | `'''` | reject | accept | a bare `'` is accepted as char content |
| F | `1e_+0` | reject | accept | `_` allowed before the exponent sign |
| G | ` #![enable(..)]` | reject | accept | leading whitespace before extensions |

Each row is a **self-checking** assertion: the transcription is faithful to the
doc, the parser behaves as claimed, and the two disagree — a wrong expectation
fails loudly instead of passing silently. Everywhere else (the `CONFORMANT`
battery, spanning every production family) the transcription and the parser agree.

Inputs ron rejects for **semantic** reasons (`911u8`, `-1u8`, `Some()`) are pinned
separately in `SEMANTIC_NOT_GRAMMAR`, so the differential does not mistake them for
conformance gaps.

## CI

A dedicated `Conformance` job runs `cd conformance && cargo test`, isolated from the
main test / MSRV matrix (`pest` is pulled into this job only).

## Note

This is the executable evidence behind #614. If useful, I'm happy to also open the
`grammar.md` fix PR; once the doc is corrected the `KNOWN_DOC_BUGS` entries fold
into the conformant battery and the divergence set goes to zero, leaving the test
as a guard against any new drift.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,6 +105,25 @@ jobs:
           override: true
       - run: cd fuzz && cargo clippy --all -- -D warnings
 
+  conformance:
+    name: "Conformance: grammar differential"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+        with:
+          key: conformance
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+      # The grammar-conformance differential lives in an isolated sub-crate (like
+      # `fuzz/`) so `pest` never enters ron's own dependency graph or MSRV job. It
+      # transcribes docs/grammar.md into a PEG acceptor and diffs it, input by
+      # input, against the parser; a divergence is a parser bug or a doc bug.
+      - run: cd conformance && cargo test -- --nocapture
+
   rustfmt:
     name: "Format: stable"
     runs-on: ubuntu-latest
--- a/conformance/.gitignore
+++ b/conformance/.gitignore
@@ -1,0 +1,2 @@
+/target
+/Cargo.lock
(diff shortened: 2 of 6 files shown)


juntyr:
@enomado Thank you for all of your PRs today! I will have a look at them over the coming days once I have some free time

